PR #4, opened 2016-03-07
Port TCK tests

ericbottard:


markfisher:
I think we should have a separate `spring-cloud-deployer-spi-test` project for this. SPI implementation projects would depend on that one with *test* scope.

spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/LocalProcessDeployerIntegrationTests.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * Integration tests for LocalProcessDeployer.
+ *
+ * @author Eric Bottard
+ */
+@SpringApplicationConfiguration(classes = LocalProcessDeployerIntegrationTests.Config.class)
+public class LocalProcessDeployerIntegrationTests extends AbstractProcessDeployerTests {
```
markfisher: if the subclass is expected to have "Integration" in its name, we probably should go ahead and rename the base class too (also the base class' JavaDoc explicitly states "integration" tests)

spring-cloud-deployer-spi/pom.xml
```diff
@@ -22,6 +22,22 @@
...
+		</dependency>
+		<dependency>
+			<groupId>junit</groupId>
+			<artifactId>junit</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>org.hamcrest</groupId>
+			<artifactId>hamcrest-all</artifactId>
+			<version>1.3</version>
```
markfisher: should move this version to a property

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/test/AbstractProcessDeployerTests.java
```diff
@@ -0,0 +1,253 @@
...
+		assertThat(status.getState(), is(unknown));
+	}
+
+	/**
+	 * Tests a simple deploy-undeploy cycle.
+	 */
+	@Test
+	public void testSimpleDeployment() {
+		AppDefinition definition = new AppDefinition(randomName(), null);
+		Resource resource = mavenResource("org.springframework.cloud.stream.module", "time-source","jar","exec","1.0.0.BUILD-SNAPSHOT");
```
markfisher: should have a space after each comma for these

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/test/AbstractProcessDeployerTests.java
```diff
@@ -0,0 +1,253 @@
...
+
+
+	/**
+	 * Represents a timeout for querying status, with repetitive queries until a certain number have been made.
+	 *
+	 * @author Eric Bottard
+	 */
+	protected static class Attempts {
+
+		public final int noAttempts;
```
markfisher: `noAttempts` is a confusing name... perhaps `maxAttempts` would be better?

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/test/AbstractProcessDeployerTests.java
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+
+
+	/**
+	 * Represents a timeout for querying status, with repetitive queries until a certain number have been made.
+	 *
+	 * @author Eric Bottard
+	 */
+	protected static class Attempts {
```
markfisher: This class name feels off, and the fact that in the comment above and in the actual var names used in the test code it's always referred to as "timeout" is confusing. That said, it seems like more than "timeout" as well since it encompasses the max number of attempts and the pause between each.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/test/AbstractProcessDeployerTests.java
```diff
@@ -0,0 +1,253 @@
...
+
+		timeout = undeploymentTimeout();
+		processDeployer().undeploy(deploymentId);
+		assertThat(deploymentId, eventually(hasStatusThat(
+				Matchers.<ProcessStatus>hasProperty("state", is(unknown))), timeout.noAttempts, timeout.pause));
+
+	}
+
+	/**
+	 * Tests that a module which takes a long time to deploy is correctly reported as deploying.
```
markfisher: we should change "module" to "app" throughout

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/test/AbstractProcessDeployerTests.java
```diff
@@ -0,0 +1,253 @@
...
+	/**
+	 * Return the timeout to use for repeatedly querying a module while it is being un-deployed.
+	 * Default value is one minute, being queried every 5 seconds.
+	 */
+	protected Attempts undeploymentTimeout() {
+		return new Attempts(20, 5000);
+	}
+
+	private Resource mavenResource(String groupId, String artifactId, String ext, String classifier, String version) {
+		// TODO: actually resolve using Aether,
```
markfisher: this could use the `MavenResource` that is part of this project (it needs some work, but currently it resolves from the local repo only)

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/test/EventuallyMatcher.java
```diff
@@ -0,0 +1,83 @@
...
+
+	private int noAttempts;
+
+	private int pause;
+
+	public EventuallyMatcher(Matcher<U> delegate) {
+		this(delegate, 20, 100);
+	}
+
+	public EventuallyMatcher(Matcher<U> delegate, int noAttempts, int pause) {
```
markfisher: I think `maxAttempts` would be more clear than `noAttempts` throughout (for one thing it IS a max... if the match occurs for any attempt less than the max it will not continue)

markfisher:
I rebased this PR branch on my fork and added one commit to address the Process->App name changes. You can grab it here:
https://github.com/markfisher/spring-cloud-deployer/commits/portTCK

I did not yet address any of the comments I made above (except I think for one "module" -> "app" replacement)... just wanted to focus on the rebase itself.

markfisher:
rebased, squashed, and merged as https://github.com/spring-cloud/spring-cloud-deployer/commit/f27a26b7ac34d1895a5579b5a4a2e84c8dffdcb9

also addressed the comments above with https://github.com/spring-cloud/spring-cloud-deployer/commit/3a899e0ac8d2c8aa9008adf8ad6a8c416da1e4da

PR #5, opened 2016-03-08
Fixes #2 - null filename in MavenResource.

### LINKED ISSUE #2: MavenResource should implement getFilename()
`AbstractResource.getFilename()` always return `null`. We should override this so that it i.e. returns `timestamp-task-1.0.0.BUILD-SNAPSHOT-exec.jar`.

venilnoronha:
Hi,

I've overridden the getFilename method from AbstractResource in MavenResource to return the actual filename. I've included a JUnit test too. This should fix #2.

Thanks,
Venil Noronha

### DIFF
--- a/spring-cloud-deployer-resource-maven/pom.xml
+++ b/spring-cloud-deployer-resource-maven/pom.xml
@@ -52,6 +52,10 @@
 			<artifactId>maven-aether-provider</artifactId>
 			<version>${maven.version}</version>
 		</dependency>
+		<dependency>
+			<groupId>junit</groupId>
+			<artifactId>junit</artifactId>
+		</dependency>
 	</dependencies>
 
 </project>
--- a/spring-cloud-deployer-resource-maven/src/test/java/org/springframework/cloud/deployer/resource/maven/MavenResourceTests.java
+++ b/spring-cloud-deployer-resource-maven/src/test/java/org/springframework/cloud/deployer/resource/maven/MavenResourceTests.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2015-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


spring-cloud-deployer-resource-maven/pom.xml
```diff
@@ -52,6 +52,10 @@
 			<artifactId>maven-aether-provider</artifactId>
 			<version>${maven.version}</version>
 		</dependency>
+		<dependency>
+			<groupId>junit</groupId>
+			<artifactId>junit</artifactId>
```
markfisher: should add 'test' scope here

spring-cloud-deployer-resource-maven/src/test/java/org/springframework/cloud/deployer/resource/maven/MavenResourceTests.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright 2015-2016 the original author or authors.
```
markfisher: 2016 alone would be more appropriate since this is new

spring-cloud-deployer-resource-maven/src/test/java/org/springframework/cloud/deployer/resource/maven/MavenResourceTests.java
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.deployer.resource.maven;
+
+import static org.junit.Assert.*;
```
markfisher: we avoid wildcard imports, and add an empty line between static and non-static imports

markfisher:
Thanks for the contribution @venilnoronha!

I can address the comments while merging.

markfisher:
addressed comments, squashed and merged as https://github.com/spring-cloud/spring-cloud-deployer/commit/07833ab89f95cef8628c52ad49f46b0385adc138

Thanks again for the contribution @venilnoronha !

venilnoronha:
Thank you @markfisher

PR #9, opened 2016-03-10
Establish LocalDeployerAutoConfiguration creation order

cppwfs:
It should be created prior to utilized prior to DataFlowControllerAutoConfiguration

This is to prevent the following exception:
[org.springframework.cloud.deployer.spi.local.LocalDeployerProperties] is defined: expected single matching bean but found 2: localDeployerProperties,deployer.local.CONFIGURATION_PROPERTIES


markfisher:
actually sorry I think I misled you by showing a config from a branch... it's not the autoconfig-before that is crucial here (that will be added in a later change), but the avoidance of duplicating the bean creation for the ConfigurationProperties. So, just removing that should be sufficient

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-deployer/commit/b82379e2d9db6e4a652c70aeecb48d0c9c7d00a1

PR #8, opened 2016-03-10
Add URI support for MavenResource

jvalkeal:
- Adding better support for defining remote repositories.
- New parse methods for URI's
  "http://my.repo.io/repo/path/org.springframework.sample:some-app:jar:exec:2.0.0"
  "maven:http://my.repo.io/repo/path/org.springframework.sample:some-app:jar:exec:2.0.0"
  "maven:org.springframework.sample:some-app:jar:exec:2.0.0"
  which automatically adds a remote repo for that artifact
- exists() returns false instead of throwing if artifact doesn't resolve.
- Fixes #6
- Fixes #7

PR #12, opened 2016-03-15
add MavenProperties to MavenResource

markfisher:


markfisher:
This is an alternative to #8 (but we should commit the `exists()` part of that PR regardless)

cc @jvalkeal

jvalkeal:
We can do that `exists()` thing with other PR.

This mostly looks OK, but you should probably copy over those tests(concept what tests do) verifying that remote maven repository settings actually works. If I delete `time` and `log` modules from my local m2 cache, this build actually fails(for same reason why CI build fails).

markfisher:
I will copy over the tests and `exists()` from your PR and update this one.

markfisher:
@jvalkeal I added `exists()` and several of your tests (with some changes)... let me know what you think

jvalkeal:
Looks good to me, lets see if CI barks.

PR #19, opened 2016-03-18
DO NOT MERGE async deployer poc

jvalkeal:
This is a little async comparison for reactor Mono/Flux vs jdk8 CompletableFuture/Stream.
- Just for review!!!
- Quick hacking, just adding new methods to SPI to see how things looks like!!!
- New methods having `1` is for reactor, methods with `2` is for jdk8.
- Imho, reactor is much _nicer_ to use, jdk8 just feels awkard.
- `CompletableFuture` does its stuff immediately and not like reactor which does it
  stuff when there's demand. So with jdk8, no backpressure support.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,8 +20,9 @@
 
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		<java.version>1.7</java.version>
+		<java.version>1.8</java.version>
 		<spring-cloud.version>Brixton.BUILD-SNAPSHOT</spring-cloud.version>
+		<reactor.version>2.5.0.M2</reactor.version>
 	</properties>
 
 	<modules>
@@ -35,6 +36,11 @@
 
 	<dependencyManagement>
 		<dependencies>
+			<dependency>
+				<groupId>io.projectreactor</groupId>
+				<artifactId>reactor-core</artifactId>
+				<version>${reactor.version}</version>
+			</dependency>
 			<dependency>
 				<groupId>org.springframework.boot</groupId>
 				<artifactId>spring-boot-starter-parent</artifactId>
@@ -102,5 +108,17 @@
 			</pluginRepositories>
 		</profile>
 	</profiles>
+
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.codehaus.mojo</groupId>
+				<artifactId>animal-sniffer-maven-plugin</artifactId>
+				<configuration>
+					<skip>true</skip>
+				</configuration>
+			</plugin>
+		</plugins>
+	</build>
 
 </project>
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
@@ -25,19 +25,25 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
 
 import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
 
+import org.reactivestreams.Publisher;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.cloud.deployer.spi.app.AppDeployer;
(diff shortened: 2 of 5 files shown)


markpollack:
From what I can see, the use of both APIs is limited in this PR since both read identically and there is a call to `get()` on the next statement in either case. The ease of use of the mono/flux API doesn't come out in this case, nor does the advantage of having backpressure.  In offline chat, I can see that for the deployer API itself to be async would have some advantages in that it can be time consuming to return from a sync deployment call (the work on the server side takes a while) so having an async api that has nice composition abilities would be an advantage.

jvalkeal:
I think one good thing to do is to open a code of `CompletableFuture` and see how it works and how you'd use it with `Stream. It's relatively awkward when you actually start to use it. `Flux/Mono` from a user perspective just feel like something `CompletableFuture` should have been from a beginning.

That's a reason, if we do an asynch SPI, I'd rather do it on our stuff than on top of jdk8. We're going to do more and more stuff on flux/mono with spring5 so things would naturally integrate more cleanly.

IMHO, it's almost impossible to explain why flux/mono feels better until you try to use `CompletableFuture` and especially try to understand how it works.

smaldini:
@jvalkeal is referring to the lazy pull property of Mono.get() which requests as you invoke it in addition to a centralized API for non-blocking/blocking with/without backpressure. In the end a Mono is a greater common denominator and can be converted toCompletableFuture() (as Flux can be converted toStream()).

The ultimate goal of a Reactive Stream type of 1 data vs CompletableFuture is to leave the data where it is at first (e.g. a database or anywhere outside the application scope) until someone requests it regardless of it being one or many results. Additionally, the thread of this new virtual "timeline" can be affected and many operators are built-in with Java 8 functional callbacks to work with the flowing result.

CompletableFuture is push based and block-pull "only", where Mono is push/pull (push if the data is not ready when pull or "request" is done). You can keep adding continuations (or operators that will run when the data will get here) or just block immediately, which should be less and less required as the end-to-end flow can get away of it.

sdeleuze:
I think we all come to these question while trying to implement real asynchronous non-blocking API, and I think there are good reasons why we are moving to Reactive APIs. I can try to explain them in addition to what @smaldini said.

The first need that we generally identify is to provide an asynchronous API/SPI in order to fix some scalability issue (threads are a limited resource) and provide a more event driven way to deal with operation that take some time to complete, most of the time due to IO.

Before Java 8 this was not obvious because callback based API are not easy to use (especially without lambda and when you need to chain calls), and the [Future](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Future.html) type is just a trap because it is maybe asynchronous but totally useless since it is blocking, the only way to get the result it to use the `get()` method that blocks the current thread.

With Java 8 bringing lambdas, `CompletableFuture` and `Stream`, I initially thought that we had everything needed to achieve this goal of providing asynchronous non-blocking API/SPI to the user. But I was wrong. Unlike `Future`, `CompletableFuture` is a real async non-blocking type designed to deal with single deferred value, but sadly [its API](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html) is a nightmare to use. That said, we could maybe think that's a price worth to pay to use only Java 8 types for our async non-blocking goal ...

... but that's where we come to another big misunderstanding about Java  8 `Stream` type. [As explain by Brian Goetz himself](http://stackoverflow.com/a/18826615/1092077), `Stream` has been designed to deal efficiently with data that can be accessed with no latency, it is not suitable to deal with IO and other kind of streams where data has some latency. Even without talking about back-pressure, that's one of the reason why you need a Reactive library (Reactor, RxJava) as soon as you want to deal with IO based streams which is the #1 use case we have in various Spring projects.

So no choice, you need to use RxJava `Observable`, Reactor `Flux` or Akka Stream when you need to deal asynchronously with streams of data. So what do you choose for async single results ? `CompletableFuture` and its strange API which will also integrate badly in you Reactive pipeline? Or RxJava `Single` / Reactor `Mono` types that provides a nice API consistent with the Reactive type you are using for N values ? I think the choice is obvious.

That's why even without talking about back-pressure (which is VERY useful IMO to deal with slow/fast producer/consumer and request only data when you are ready to process it) I think it makes sense to use a Reactive library to perform async non-blocking processing and expose an API that allows the user to deal with that kind of operation in a maintanable way. That's why @nebhale chose that for Cloud Foundry Java client, and why we are in the process of providing a Reactive API and implementation on various part of the Spring portfolio where we are dealing with IO and/or latency, including [`spring-cloud-stream`](https://github.com/spring-cloud/spring-cloud-stream/issues/458).

I think `spring-cloud-deployer` also fits in this category and could take advantage of providing such Reactive API/SPI.

markpollack:
guys, I get it, as per my original comment.   I just don't know when we can do it.  ATM there are more basic things to take care of across the 15 or so projects we are trying to release very soon.

smaldini:
@markpollack Agree that the point is not to overwhelm you with this extra dimension to take care of. We are in a similar situation and we need to move and help moving bit by bit about all the major spring projects across our extensive portfolio :+1: If someone is willing to offer early on support of Reactive like in the PR, even bridged, that's a benediction. Additionally we want to tune this common reactive utils that is reactor-core to fulfill missions for Cloud too and the more feedback we can catch the better :dancer:

markpollack:
Closing as there hasn't been much demand for this and can revisit the closed PR should interest arise again.

PR #22, opened 2016-03-22
Infer count property from the environment

ilayaperumalg:
- If environment property doesn't have value for `spring.cloud.deployer.count` then check if the definition property has it
 - If both don't have it set, then use the default count 1


markfisher:
closing this since count should *only* be an environment property... the properties in the `AppDefinition` are for passing as boot properties to an application only

markfisher:
merged along with this change: https://github.com/spring-cloud/spring-cloud-deployer/commit/ec2a840690528785948a469c77f7eed4f28832cc

Issue #11, opened 2016-03-10
Aether is not thread safe


jvalkeal:
"Out of the box, Aether provides nothing that helps to deal with concurrent access to a single local repository."
https://dev.eclipse.org/mhonarc/lists/aether-users/msg00312.html

We need to provide some solution to this. Above email thread links to https://github.com/takari/takari-local-repository.

jvalkeal:
Not 100% if this trace relates to this:
```
16-06-28T10:55:49.59+0200 [APP/0]      OUT Caused by: java.io.FileNotFoundException: /home/vcap/.m2/repository/org/springframework/cloud/stream/app/twitterstream-source-rabbit/1.0.0.BUILD-SNAPSHOT/twitterstream-source-rabbit-1.0.0.BUILD-SNAPSHOT.jar.part.lock (No such file or directory)
2016-06-28T10:55:49.59+0200 [APP/0]      OUT     at java.io.RandomAccessFile.open0(Native Method) ~[na:1.8.0_91]
2016-06-28T10:55:49.59+0200 [APP/0]      OUT     at java.io.RandomAccessFile.open(RandomAccessFile.java:316) ~[na:1.8.0_91]
2016-06-28T10:55:49.59+0200 [APP/0]      OUT     at java.io.RandomAccessFile.<init>(RandomAccessFile.java:243) ~[na:1.8.0_91]
2016-06-28T10:55:49.59+0200 [APP/0]      OUT     at org.eclipse.aether.connector.basic.PartialFile$LockFile.tryLock(PartialFile.java:123) ~[aether-connector-basic-1.0.2.v20150114.jar!/:na]
2016-06-28T10:55:49.59+0200 [APP/0]      OUT     at org.eclipse.aether.connector.basic.PartialFile$LockFile.lock(PartialFile.java:71) ~[aether-connector-basic-1.0.2.v20150114.jar!/:na]
2016-06-28T10:55:49.59+0200 [APP/0]      OUT     at org.eclipse.aether.connector.basic.PartialFile$LockFile.<init>(PartialFile.java:58) ~[aether-connector-basic-1.0.2.v20150114.jar!/:na]
2016-06-28T10:55:49.59+0200 [APP/0]      OUT     at org.eclipse.aether.connector.basic.PartialFile$Factory.newInstance(PartialFile.java:210) ~[aether-connector-basic-1.0.2.v20150114.jar!/:na]
[...]

PR #23, opened 2016-03-22
add MavenResourceLoader

markfisher:
Resolves spring-cloud/spring-cloud-dataflow#403

### DIFF
--- a/spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenResourceLoader.java
+++ b/spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenResourceLoader.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.deployer.resource.maven;
+
+import org.springframework.core.io.Resource;
+import org.springframework.core.io.ResourceLoader;
+import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
+
+/**
+ * A {@link ResourceLoader} that loads {@link MavenResource}s from locations of the format
+ * {@literal maven://<coordinates>} where the value for "coordinates" conforms to the rules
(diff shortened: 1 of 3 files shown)


markfisher:
maybe the resource loader should inject the `MavenArtifactResolver` into each `MavenResource` so that can be a shared bean (as opposed to only providing `MavenProperties` and having each `MavenResource` create its own `MavenArtifactResolver`).... then again, this should be considered in light of the thread-safety concerns. Even if the current approach is heavier, is it safer from that perspective @jvalkeal ?

jvalkeal:
@markfisher I'm nervous about sharing `MavenArtifactResolver` because of thread safety #11

jvalkeal:
1. How we're planning to configure currently used `ResourceLoader`'s.
2. As we want to support more types like, `http`, `s3`, `hdfs`, etc, I believe we need to have a some sort of a composite `ResourceLoader` which dispatch into a correct one and possibly defines a default if resource prefix is missing.
3. Actual server impl, need to be able to define their own supported loaders, as yarn would definitely use `hdfs` while pcf probably would not(unless we really want to add spring hadoop deps there). Or maybe at some day on pcf there is a blob store which only works within pcf and we want to create a loader for that.
4. Also need to make sure we don't get into trouble with `ApplicationContext` which is also a `ResourceLoader`. We got into some trouble with `HdfsResourceLoader`.

spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenResourceLoader.java
```diff
@@ -0,0 +1,73 @@
...
+	 * Returns a {@link MavenResource} for the provided location.
+	 *
+	 * @param location the coordinates conforming to the rules described on
+	 * {@link MavenResource#parse(String)}. May optionally be preceded by {@value #URI_SCHEME}
+	 * followed by a colon and zero or more forward slashes, e.g.
+	 * {@literal maven://group:artifact:version}
+	 * @return the {@link MavenResource}
+	 */
+	@Override
+	public Resource getResource(String location) {
```
jvalkeal: I'd follow concepts from org.springframework.core.io.DefaultResourceLoader more closely how prefix(whether its correct, missing or malformed/unsupported) is used. I'd make sure that given prefix is `maven://` before trying to use it, if no prefix use it as is, if wrong throw `Exception` of some sort. Thought `ResourceLoader.getResource()` is a bit vague how it should work in error cases.

markfisher: But... `DefaultResourceLoader` has to be strict since it handles many different variations (classpath, url, file), and the one that is relevant is determined by the prefix (or a simple '/' in the first position)... whereas in this case, we're always going to create a `MavenResource` and that prefix doesn't provide any value (not at this level, it would provide value at the `DelegatingResourceLoader` level though). So, I was thinking we might as well be lenient.

jvalkeal:
@markfisher u plan to make any changes or shall I merge? Think my questions were pretty much answered in that google doc.

markfisher:
go ahead and merge if you're ok with it - I'm sure it will continue to evolve

PR #33, opened 2016-03-31
Adding Docker resource sub-project

trisberg:
- Adding DockerResourceLoader and DockerResource

### DIFF
--- a/spring-cloud-deployer-resource-docker/src/main/java/org/springframework/cloud/deployer/resource/docker/DockerResource.java
+++ b/spring-cloud-deployer-resource-docker/src/main/java/org/springframework/cloud/deployer/resource/docker/DockerResource.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.deployer.resource.docker;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+
+import org.springframework.core.io.AbstractResource;
+import org.springframework.core.io.Resource;
+import org.springframework.util.Assert;
+
+/**
+ * A {@link Resource} implementation for resolving a Docker image.
+ *
+ * Note: {@link #getInputStream()} throws {@code UnsupportedOperationException}.
+ *
+ * @author Thomas Risberg
+ */
+public class DockerResource extends AbstractResource {
+
+	public static String URI_SCHEME = "docker";
+
+	private String scheme = "docker";
+
+	private URI uri;
+
+	/**
+	 * Create a new {@code DockerResource} from an image name.
+	 * @param imageName the name of the image in a docker registry.
+	 */
+	public DockerResource(String imageName) {
+		this.uri = URI.create(URI_SCHEME + ":" + imageName);
+	}
+
+	/**
+	 * Create a new {@code DockerResource} from a URI
+	 * @param uri a URI
+	 */
+	public DockerResource(URI uri) {
+		Assert.isTrue("docker".equals(uri.getScheme()), "A 'docker' scheme is required");
+		this.uri = uri;
+	}
+
+
+	@Override
+	public String getDescription() {
+		return "Docker Resource [" + uri + "]";
+	}
+
+	/**
+	 * This implementation currently throws {@code UnsupportedOperationException}
+	 */
+	@Override
+	public InputStream getInputStream() throws IOException {
+		throw new UnsupportedOperationException("getInputStream not supported");
+	}
+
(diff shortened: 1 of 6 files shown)


spring-cloud-deployer-resource-docker/src/main/java/org/springframework/cloud/deployer/resource/docker/DockerResourceLoader.java
```diff
@@ -0,0 +1,57 @@
...
+	 * Returns a {@link DockerResource} for the provided location.
+	 *
+	 * @param location the image location. May optionally be preceded by {@value DockerResource#URI_SCHEME}
+	 * followed by a colon, e.g. {@literal docker:springcloud/app-name:tag}
+	 * @return the {@link DockerResource}
+	 */
+	@Override
+	public Resource getResource(String location) {
+		Assert.hasText(location, "image location is required");
+		String image = location.replaceFirst(DockerResource.URI_SCHEME + ":", "");
```
markfisher: would "docker://" be allowed as well? if so, the regex could be: `DockerResource.URI_SCHEME + ":\\/*"`

spring-cloud-deployer-resource-docker/src/main/java/org/springframework/cloud/deployer/resource/docker/DockerResource.java
```diff
@@ -0,0 +1,78 @@
...
+	 */
+	public DockerResource(String imageName) {
+		this.uri = URI.create(URI_SCHEME + ":" + imageName);
+	}
+
+	/**
+	 * Create a new {@code DockerResource} from a URI
+	 * @param uri a URI
+	 */
+	public DockerResource(URI uri) {
```
markfisher: we should probably `Assert.notNull(uri)` also

spring-cloud-deployer-resource-docker/src/main/java/org/springframework/cloud/deployer/resource/docker/DockerResource.java
```diff
@@ -0,0 +1,78 @@
...
+
+	private String scheme = "docker";
+
+	private URI uri;
+
+	/**
+	 * Create a new {@code DockerResource} from an image name.
+	 * @param imageName the name of the image in a docker registry.
+	 */
+	public DockerResource(String imageName) {
```
markfisher: should add an `Assert.hasText(imageName)`

spring-cloud-deployer-resource-docker/src/main/java/org/springframework/cloud/deployer/resource/docker/DockerResource.java
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * Note: {@link #getInputStream()} throws {@code UnsupportedOperationException}.
+ *
+ * @author Thomas Risberg
+ */
+public class DockerResource extends AbstractResource {
+
+	public static String URI_SCHEME = "docker";
+
+	private String scheme = "docker";
```
markfisher: might as well use the constant here ;)

spring-cloud-deployer-resource-docker/src/main/java/org/springframework/cloud/deployer/resource/docker/DockerResourceLoader.java
```diff
@@ -0,0 +1,57 @@
...
+import org.springframework.util.ClassUtils;
+
+/**
+ * A {@link ResourceLoader} that loads {@link DockerResource}s from locations of the format
+ * {@literal docker:<repository>:<tag>} where the value for "repository:tag" conforms to the rules
+ * for referencing Docker images.
+ *
+ * @author Thomas Risberg
+ */
+public class DockerResourceLoader  implements ResourceLoader {
```
markfisher: would this eventually include information about one or more particular registries?... and auth credentials if necessary 

assuming we eventually implement the resource's `getInputStream()`, we would need to pass those properties along to the `DockerResource` that is loaded - similar to how `MavenProperties` are passed to a `MavenResource` from the `MavenResourceLoader`... and whereas the Maven implementation uses the Aether library, we would use the Docker Java API

spring-cloud-deployer-resource-docker/src/main/java/org/springframework/cloud/deployer/resource/docker/DockerResourceLoader.java
trisberg: This is our own scheme so not sure what // would add to it, but I guess we could remove it if provided

spring-cloud-deployer-resource-docker/src/main/java/org/springframework/cloud/deployer/resource/docker/DockerResource.java
trisberg: oops

spring-cloud-deployer-resource-docker/src/main/java/org/springframework/cloud/deployer/resource/docker/DockerResourceLoader.java
trisberg: Yes, we would need to be able to handle credentials, might have to pass them along for deployments. What we have now is bare-bones but gets the work done for now.

markfisher: sounds good

markfisher:
LGTM; merging

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-deployer/commit/598c1902d75c785a3bfeeecdd8f8f6d9bce06dc5

PR #35, opened 2016-04-07
[For discussion] Basic debug support for local deployer

aclement:
Created two pull requests (one here on the deployer, the other on dataflow), these are for github issue https://github.com/spring-cloud/spring-cloud-dataflow/issues/369 "Add debug capabilities to LocalAppDeployer".

With these two in place you can turn on debug by specifying "module.foo.debug-port=8800" as a deployment property.

The first pull request (on the deployer) recognizes a debug-port property and uses it to add a debug agent to the configuration for the Process it launches. 

The second pull request (on dataflow) recognizes the deployment pattern module.<xxx>.debug-port=XXX and maps that to the debug-port property in the deployer.

This works but are the properties in the right place? Is it ok to have this as a deployment config property? (It is quite nice that turning debug on/off does not require meddling with the stream definition). What do we do for other deployers? Do we need a simpler 'debug=true/false' option alongside this for any that don't need a port, or do we need a deployer specific namespace subsection for this?


markpollack:
This has been superseded by https://github.com/spring-cloud/spring-cloud-deployer-local/pull/85

PR #36, opened 2016-04-08
Add option for command line args

### LINKED ISSUE #20: Add option for command line args
As a developer, I want to be able to configure command line args as a separate set of properties to be passed to an app when it's deployed.

Backstory
When running something like a task or Spring Batch job, there are properties I'll want to configure once per app definition (database connection information for example), and yet there are items that I will want to be different with each launch (for example processing date).  I'd expect the current properties value in an AppDefinition to address the first set of values (database properties), but not the second (especially within the SCDF environment).  This story is to create a subclass of AppLaunchRequest that has an additional set of properties for tasks that will be passed to the boot jar as command line args.

jvalkeal:
- AppDeploymentRequest now has commandlineArguments
  as Set.
- Small refactoring by adding shared AbstractDeployerSupport
  for LocalAppDeployer and LocalTaskLauncher.
- Only supports arbitrary java arguments, not options which
  should go between `java` and `-jar`. Arguments are placed after
  `java -jar awesomefat.jar` as is.
- We don't have a proper way to test argument passing
  as we can't read stdout/stderr in tests, otherwise
  using i.e. timestamp task would be easy to verify
  correct passing via its `format` option.
- Actual integration in a dataflow side will be done
  in github.com/spring-cloud/spring-cloud-dataflow/issues/503.
- Fixes #20

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractDeployerSupport.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractDeployerSupport.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.deployer.spi.local;
+
+import java.util.ArrayList;
+
+import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
+import org.springframework.util.Assert;
+
+/**
+ * Base class for app deployer and task launcher providing
+ * support for common functionality.
+ *
+ * @author Janne Valkealahti
+ */
+public abstract class AbstractDeployerSupport {
+
+	private final LocalDeployerProperties properties;
+
+	/**
+	 * Instantiates a new abstract deployer support.
+	 *
+	 * @param properties the local deployer properties
+	 */
+	public AbstractDeployerSupport(LocalDeployerProperties properties) {
+		Assert.notNull(properties, "LocalDeployerProperties must not be null");
+		this.properties = properties;
+	}
+
+	/**
+	 * Gets the local deployer properties.
+	 *
+	 * @return the local deployer properties
+	 */
+	protected LocalDeployerProperties getLocalDeployerProperties() {
+		return properties;
+	}
+
+	/**
+	 * Builds the jar execution command.
+	 *
+	 * @param jarPath the jar path
+	 * @param request the request
+	 * @return the string[]
+	 */
+	protected String[] buildJarExecutionCommand(String jarPath, AppDeploymentRequest request) {
+		ArrayList<String> commands = new ArrayList<String>();
+		commands.add(properties.getJavaCmd());
+		commands.add("-Dfile.encoding=UTF-8");
+		commands.add("-jar");
+		commands.add(jarPath);
+		commands.addAll(request.getCommandlineArguments());
+		return commands.toArray(new String[0]);
+	}
+}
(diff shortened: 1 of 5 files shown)


spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/core/AppDeploymentRequest.java
```diff
@@ -65,13 +74,29 @@
...
+	/**
+	 * Construct an {@code AppDeploymentRequest}.
+	 *
+	 * @param definition app definition
+	 * @param resource resource for the underlying app's artifact
+	 * @param environmentProperties map of environment properties; may be {@code null}
+	 * @param commandlineArguments set of command line arguments; may be {@code null}
+	 */
+	public AppDeploymentRequest(AppDefinition definition, Resource resource,
+			Map<String, String> environmentProperties, Set<String> commandlineArguments) {
```
trisberg: Why a Set? Wouldn't List be better to guarantee ordering and allowing duplicate entries?

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractDeployerSupport.java
```diff
@@ -0,0 +1,69 @@
...
+		Assert.notNull(properties, "LocalDeployerProperties must not be null");
+		this.properties = properties;
+	}
+
+	/**
+	 * Gets the local deployer properties.
+	 *
+	 * @return the local deployer properties
+	 */
+	protected LocalDeployerProperties getLocalDeployerProperties() {
```
mminella: This should be final if it's going to be called from subclasses to prevent subclasses from overriding.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/core/AppDeploymentRequest.java
jvalkeal: Well I picked set not to have duplicate entries, why allowing dups? Though even set really doesn't guarantee that if one entry has multiple options.

mminella: +1

jvalkeal: But yeah, maybe List is better indeed.

mminella:
While you can't verify that the parameter is used that way, we could have a task that requires a parameter (aka will fail if it's not there) for testing.  Right now, the test here really doesn't do anything other than a compile check since the app being requested does nothing with the parameter...

jvalkeal:
There's similar issues with other tests and eric created `integration-test-processor` for that which use is a bit pita but kinda understandable from spi tests point of view. This we should try to create some hooks to local deployer so that it would expose its logging directory as then testing would be easy. With yarn minicluster tests I have access to container logs where I mostly check from stdout/stderr that stuff works.

mminella:
I'm just thinking that having a module that validated what was passed in would suffice.  If it didn't get what was expected, it fails...if it does it runs.  That test module isn't even a part of this project which is also problematic (since if you want to update it to cover a test scenario, you can't).

jvalkeal:
I really hate idea of creating a dummy test modules and publishing those to repos, but probably best shot in a short term. Maybe one for tasks and then we could verify if tasks run ok.

mminella:
Wouldn't the local repo work for the publishing aspect?  If not,then I agree

jvalkeal:
Not sure what you mean by that comment. Nothing is published into local repo during a build/test. Thought if in a resolver we'd support `file://` resource then we might be able to build jar and use that for deploy tests.

mminella:
If you do a maven install it ends up in your local repo and that would be available for the test, right?

jvalkeal:
I don't understand a topic around deploying to local maven repo as none will ever happen during a normal build/test cycle?

mminella:
A "normal build/test cycle" for me is `$ mvn clean install`...is that not normal?

jvalkeal:
`mvn clean package` is more normal, but it doesn't matter. when test phase is run it's always executed before installing anything to local repo.

mminella:
I'm trying to merge this but I'm getting test failures.  Specifically: `LocalAppDeployerIntegrationTests#testArgumentsPassing`, `LocalAppDeployerIntegrationTests#testRedeploy`, `LocalAppDeployerIntegrationTests#testParameterPassing`, `LocalAppDeployerIntegrationTests#testSimpleDeployment`.  All of them have the same error:

```
testArgumentsPassing(org.springframework.cloud.deployer.spi.local.LocalAppDeployerIntegrationTests)  Time elapsed: 120.3 sec  <<< FAILURE!
java.lang.AssertionError: 
Expected: hasProperty("state", is <deployed>), trying at most 12 times
     but: failed after 12*5000=60000ms:
status of "null.a304b8a6-c51b-42af-ac55-b40e4d22d5ea" property 'state' was <undeployed>, status of "null.a304b8a6-c51b-42af-ac55-b40e4d22d5ea" property 'state' was <undeployed>, status of "null.a304b8a6-c51b-42af-ac55-b40e4d22d5ea" property 'state' was <undeployed>, status of "null.
```

jvalkeal:
closing for new PR

PR #40, opened 2016-04-12
Add cache to DelegatingResourceLoader

### LINKED ISSUE #38: Resolving stream app artifact from http endpoint fails
As a user, I registered a stream application accessible at an `http` endpoint; however, when I use it in a stream, it fails with an error.

**registration:**
```
module register --name test-log --uri http://www.dropbox.com/<LOCATION>/log-sink-1.0.0.BUILD-SNAPSHOT-exec.jar --type sink
```

**stream:**
```
stream create test --definition "time | test-log" --deploy
```

**error:**
```
Command failed org.springframework.cloud.dataflow.rest.client.DataFlowClientException: java.io.FileNotFoundException: URL [http://www.dropbox.com/<LOCATION>/log-sink-1.0.0.BUILD-SNAPSHOT-exec.jar] cannot be resolved to absolute file path because it does not reside in the file system: http://www.dropbox.com/<LOCATION>/log-sink-1.0.0.BUILD-SNAPSHOT-exec.jar

java.io.FileNotFoundException: URL [http://www.dropbox.com/<LOCATION>/log-sink-1.0.0.BUILD-SNAPSHOT-exec.jar] cannot be resolved to absolute file path because it does not reside in the file system: http://www.dropbox.com/<LOCATION>/log-sink-1.0.0.BUILD-SNAPSHOT-exec.jar

    at org.springframework.cloud.dataflow.rest.client.VndErrorResponseErrorHandler.handleError(VndErrorResponseErrorHandler.java:52)
    at org.springframework.web.client.RestTemplate.handleResponse(RestTemplate.java:641)
    at org.springframework.web.client.RestTemplate.doExecute(RestTemplate.java:597)
    at org.springframework.web.client.RestTemplate.execute(RestTemplate.java:557)
    at org.springframework.web.client.RestTemplate.postForObject(RestTemplate.java:357)
    at org.springframework.cloud.dataflow.rest.client.StreamTemplate.createStream(StreamTemplate.java:83)
    at org.springframework.cloud.dataflow.shell.command.StreamCommands.createStream(StreamCommands.java:99)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:483)
    at org.springframework.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:216)
    at org.springframework.shell.core.SimpleExecutionStrategy.invoke(SimpleExecutionStrategy.java:64)
    at org.springframework.shell.core.SimpleExecutionStrategy.execute(SimpleExecutionStrategy.java:57)
    at org.springframework.shell.core.AbstractShell.executeCommand(AbstractShell.java:130)
    at org.springframework.shell.core.JLineShell.promptLoop(JLineShell.java:533)
    at org.springframework.shell.core.JLineShell.run(JLineShell.java:179)
    at java.lang.Thread.run(Thread.java:745)
```

### LINKED ISSUE #39: Support getFile() for all resources
`Resource.getFile()` doesn't work for all resource types as for example with `UrlResource` which points to `http` you're expected to use inputstream. The one using `Resource` would not know and anyway requirement is to get access to actual physical file so I think we should transparently support a caching so that `getFile()` works.

This requirement is coming from https://github.com/spring-cloud/spring-cloud-deployer/issues/38. While user could first try to use File, then fallback to inputstream and create a tmp file, it's best to handle this transparently in a deployer core classes where we keep `Resource` support classes. We just need to make a cache directory, possibly outside of system tmp dir, so that in a perfect world file(s) are only downloaded once.

jvalkeal:
- Add a relatively naive caching into a loader
  so that resources which are not directly available
  as File can be 'cached' from its inputstream and
  then used there as a direct FileSystemResource.
- Cache key is a simple resource filename and thus
  assumes that different resources having same file
  name are identical.
- On default a cache directory is created for every
  instance of a DelegatingResourceLoader, but can be
  manually set via constructor which is something what
  dataflow server can expose as configurable option.
- Fixes #38
- Fixes #39

### DIFF
--- a/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
+++ b/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
@@ -16,31 +16,67 @@
 
 package org.springframework.cloud.deployer.resource.support;
 
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Collections;
 import java.util.Map;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.core.io.FileSystemResource;
 import org.springframework.core.io.Resource;
 import org.springframework.core.io.ResourceLoader;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
+import org.springframework.util.FileCopyUtils;
 
 /**
  * A {@link ResourceLoader} implementation that delegates to other {@link ResourceLoader} instances
  * that are stored in a Map with their associated URI schemes as the keys.
  *
+ * This implementation is also caching remote resources which are not directly accessible
+ * as {@link File} into either a given cache directory or on default a temporary location
+ * prefixed by "dataflow-resource-cache".
+ *
  * @author Mark Fisher
+ * @author Janne Valkealahti
  */
 public class DelegatingResourceLoader implements ResourceLoader {
+
+	private static final Logger logger = LoggerFactory.getLogger(DelegatingResourceLoader.class);
 
 	private final ClassLoader classLoader = ClassUtils.getDefaultClassLoader();
 
 	private final Map<String, ResourceLoader> loaders;
 
+	private final File cacheDirectory;
+
+	private final static String DEFAULT_CACHE_PREFIX = "dataflow-resource-cache";
+
+	/**
+	 * Instantiates a new delegating resource loader.
+	 *
+	 * @param loaders the loaders
+	 */
 	public DelegatingResourceLoader(Map<String, ResourceLoader> loaders) {
+		this(loaders, null);
+	}
+
+	/**
+	 * Instantiates a new delegating resource loader.
+	 *
+	 * @param loaders the loaders
+	 * @param cacheDirectory the cache directory
+	 */
+	public DelegatingResourceLoader(Map<String, ResourceLoader> loaders, File cacheDirectory) {
 		Assert.notEmpty(loaders, "at least one ResourceLoader is required");
 		this.loaders = Collections.unmodifiableMap(loaders);
+		this.cacheDirectory = initCacheDirectory(cacheDirectory);
 	}
 
 	@Override
@@ -51,9 +87,24 @@ public class DelegatingResourceLoader implements ResourceLoader {
 			Assert.notNull(scheme, "a scheme (prefix) is required");
 			ResourceLoader loader = this.loaders.get(scheme);
 			Assert.notNull(loader, String.format("no ResourceLoader available for scheme: %s", scheme));
-			return loader.getResource(location);
+
+			Resource resource = loader.getResource(location);
+			if (existsAsFile(resource)) {
+				return resource;
+			} else {
+				File cachedResource = new File(cacheDirectory, resource.getFilename());
+				if (!cachedResource.exists()) {
+					logger.info("Caching file {} as given location {}", cachedResource, location);
+					FileCopyUtils.copy(resource.getInputStream(), new FileOutputStream(cachedResource));
+				} else {
+					logger.info("Reusing cached file {} as given location {}", cachedResource, location);
+				}
+				return new FileSystemResource(cachedResource);
+			}
 		}
 		catch (URISyntaxException e) {
+			throw new IllegalArgumentException(e);
+		} catch (IOException e) {
 			throw new IllegalArgumentException(e);
 		}
 	}
@@ -63,4 +114,38 @@ public class DelegatingResourceLoader implements ResourceLoader {
 		return this.classLoader;
 	}
 
+	/**
+	 * Handles init operation of a local cache directory.
+	 *
+	 * @param cacheDirectory the cache directory
+	 * @return the directory
+	 */
+	private File initCacheDirectory(File cacheDirectory) {
+		try {
+			if (cacheDirectory == null) {
+				Path tempDirectory = Files.createTempDirectory(DEFAULT_CACHE_PREFIX);
+				return tempDirectory.toFile();
+			} else {
+				cacheDirectory.mkdirs();
+				return cacheDirectory;
+			}
+		} catch (IOException e) {
+			throw new IllegalArgumentException("Unable to create cache directory", e);
+		}
+	}
+
+	/**
+	 * Check if a resource exists as a file.
+	 *
+	 * @param resource the resource
+	 * @return true, if resource can be accessed as file
+	 */
+	private static boolean existsAsFile(Resource resource) {
+		try {
+			resource.getFile();
+			return true;
+		} catch (IOException e) {
+		}
+		return false;
+	}
 }
--- a/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
+++ b/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
@@ -28,6 +28,7 @@ import java.util.Map;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
 import org.springframework.core.io.FileSystemResource;
 import org.springframework.core.io.Resource;
 import org.springframework.core.io.ResourceLoader;
@@ -91,12 +92,14 @@ public class DelegatingResourceLoader implements ResourceLoader {
 			Resource resource = loader.getResource(location);
 			if (existsAsFile(resource)) {
 				return resource;
-			} else {
+			}
+			else {
 				File cachedResource = new File(cacheDirectory, resource.getFilename());
 				if (!cachedResource.exists()) {
 					logger.info("Caching file {} as given location {}", cachedResource, location);
 					FileCopyUtils.copy(resource.getInputStream(), new FileOutputStream(cachedResource));
-				} else {
+				}
+				else {
 					logger.info("Reusing cached file {} as given location {}", cachedResource, location);
 				}
 				return new FileSystemResource(cachedResource);
@@ -104,8 +107,9 @@ public class DelegatingResourceLoader implements ResourceLoader {
 		}
 		catch (URISyntaxException e) {
 			throw new IllegalArgumentException(e);
-		} catch (IOException e) {
-			throw new IllegalArgumentException(e);
+		}
+		catch (IOException e) {
+			throw new IllegalStateException(e);
 		}
 	}
 
@@ -125,11 +129,13 @@ public class DelegatingResourceLoader implements ResourceLoader {
 			if (cacheDirectory == null) {
 				Path tempDirectory = Files.createTempDirectory(DEFAULT_CACHE_PREFIX);
 				return tempDirectory.toFile();
-			} else {
+			}
+			else {
 				cacheDirectory.mkdirs();
 				return cacheDirectory;
 			}
-		} catch (IOException e) {
+		}
+		catch (IOException e) {
 			throw new IllegalArgumentException("Unable to create cache directory", e);
 		}
 	}
@@ -144,7 +150,8 @@ public class DelegatingResourceLoader implements ResourceLoader {
 		try {
 			resource.getFile();
 			return true;
-		} catch (IOException e) {
+		}
+		catch (IOException e) {
 		}
 		return false;
 	}
--- a/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
+++ b/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
@@ -41,7 +41,7 @@ import org.springframework.util.FileCopyUtils;
  *
  * This implementation is also caching remote resources which are not directly accessible
  * as {@link File} into either a given cache directory or on default a temporary location
- * prefixed by "dataflow-resource-cache".
+ * prefixed by "deployer-resource-cache".
  *
  * @author Mark Fisher
  * @author Janne Valkealahti
@@ -56,7 +56,7 @@ public class DelegatingResourceLoader implements ResourceLoader {
 
 	private final File cacheDirectory;
 
-	private final static String DEFAULT_CACHE_PREFIX = "dataflow-resource-cache";
+	private final static String DEFAULT_CACHE_PREFIX = "deployer-resource-cache";
 
 	/**
 	 * Instantiates a new delegating resource loader.
--- a/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
+++ b/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
@@ -94,7 +94,8 @@ public class DelegatingResourceLoader implements ResourceLoader {
 				return resource;
 			}
 			else {
-				File cachedResource = new File(cacheDirectory, resource.getFilename());
+				String cacheName = scheme + "-" + ShaUtils.sha1(location) + "-" + resource.getFilename();
+				File cachedResource = new File(cacheDirectory, cacheName);
 				if (!cachedResource.exists()) {
 					logger.info("Caching file {} as given location {}", cachedResource, location);
 					FileCopyUtils.copy(resource.getInputStream(), new FileOutputStream(cachedResource));
--- a/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/ShaUtils.java
+++ b/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/ShaUtils.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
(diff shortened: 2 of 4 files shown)


spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
```diff
@@ -16,31 +16,67 @@
...
 
+	private static final Logger logger = LoggerFactory.getLogger(DelegatingResourceLoader.class);
+
 	private final ClassLoader classLoader = ClassUtils.getDefaultClassLoader();
 
 	private final Map<String, ResourceLoader> loaders;
 
+	private final File cacheDirectory;
+
+	private final static String DEFAULT_CACHE_PREFIX = "dataflow-resource-cache";
```
markfisher: should not use "dataflow" here

jvalkeal: yeah, good catch, fixed.

spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
```diff
@@ -51,10 +87,25 @@ public Resource getResource(String location) {
...
+				} else {
+					logger.info("Reusing cached file {} as given location {}", cachedResource, location);
+				}
+				return new FileSystemResource(cachedResource);
+			}
 		}
 		catch (URISyntaxException e) {
 			throw new IllegalArgumentException(e);
+		} catch (IOException e) {
+			throw new IllegalArgumentException(e);
```
markfisher: should this throw `IllegalStateException` instead?

spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
```diff
@@ -63,4 +114,38 @@ public ClassLoader getClassLoader() {
...
+	 * Check if a resource exists as a file.
+	 *
+	 * @param resource the resource
+	 * @return true, if resource can be accessed as file
+	 */
+	private static boolean existsAsFile(Resource resource) {
+		try {
+			resource.getFile();
+			return true;
+		} catch (IOException e) {
```
markfisher: nitpick: `return false` could go in the catch block

spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
```diff
@@ -16,31 +16,67 @@
...
+import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Collections;
 import java.util.Map;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
markfisher: should be a blank line before `org.springframework` imports

spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
```diff
@@ -51,10 +87,25 @@ public Resource getResource(String location) {
 			Assert.notNull(scheme, "a scheme (prefix) is required");
 			ResourceLoader loader = this.loaders.get(scheme);
 			Assert.notNull(loader, String.format("no ResourceLoader available for scheme: %s", scheme));
-			return loader.getResource(location);
+
+			Resource resource = loader.getResource(location);
+			if (existsAsFile(resource)) {
+				return resource;
+			} else {
```
markfisher: `else` on new line

spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
```diff
@@ -51,10 +87,25 @@ public Resource getResource(String location) {
...
+				File cachedResource = new File(cacheDirectory, resource.getFilename());
+				if (!cachedResource.exists()) {
+					logger.info("Caching file {} as given location {}", cachedResource, location);
+					FileCopyUtils.copy(resource.getInputStream(), new FileOutputStream(cachedResource));
+				} else {
```
markfisher: `else` on new line

spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
```diff
@@ -51,10 +87,25 @@ public Resource getResource(String location) {
...
+					logger.info("Reusing cached file {} as given location {}", cachedResource, location);
+				}
+				return new FileSystemResource(cachedResource);
+			}
 		}
 		catch (URISyntaxException e) {
 			throw new IllegalArgumentException(e);
+		} catch (IOException e) {
```
markfisher: `catch` on new line

spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
```diff
@@ -63,4 +114,38 @@ public ClassLoader getClassLoader() {
...
+	 *
+	 * @param cacheDirectory the cache directory
+	 * @return the directory
+	 */
+	private File initCacheDirectory(File cacheDirectory) {
+		try {
+			if (cacheDirectory == null) {
+				Path tempDirectory = Files.createTempDirectory(DEFAULT_CACHE_PREFIX);
+				return tempDirectory.toFile();
+			} else {
```
markfisher: `else` on new line

spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
```diff
@@ -63,4 +114,38 @@ public ClassLoader getClassLoader() {
...
+				cacheDirectory.mkdirs();
+				return cacheDirectory;
+			}
+		} catch (IOException e) {
```
markfisher: `catch` on new line

spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
markfisher: `catch` on new line

spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
```diff
@@ -51,10 +87,25 @@ public Resource getResource(String location) {
 			Assert.notNull(scheme, "a scheme (prefix) is required");
 			ResourceLoader loader = this.loaders.get(scheme);
 			Assert.notNull(loader, String.format("no ResourceLoader available for scheme: %s", scheme));
-			return loader.getResource(location);
+
+			Resource resource = loader.getResource(location);
+			if (existsAsFile(resource)) {
+				return resource;
+			} else {
+				File cachedResource = new File(cacheDirectory, resource.getFilename());
```
ericbottard: This has several problems IMHO:
* Same name files, but with different paths will collide (_e.g._ http://foo/bar/somefile.jar and http://woot/other/somefile.jar)
* What if the resource changes on the remote end?

markfisher:
good points @ericbottard - I believe we could address both by adding a UUID to the file name... it would no longer be a "cache" (no need to check if the file exists yet), but rather copying from input stream to file would happen each time the resource is requested, and the file written would be scoped to that request

ericbottard:
> and the file written would be scoped to that request

Which could be dramatic, at least with the current implementation of metadata retrieval, _e.g._ for TAB completion.

This is a non trivial issue, really IMO.

markfisher:
maybe completion needs its own caching layer.... just a thought

ericbottard:
I think that is the best thing to do, yes. As we absolutely don't want to aggressively cache too early. Although `http` should be rarely used.

Also, I believe we should check which path Aether gives us when we request SNAPSHOTs (hoping it is a timestamped one, and not `foo-SNAPSHOT.jar`)

ericbottard:
> Also, I believe we should check which path Aether gives us when we request SNAPSHOTs (hoping it is a timestamped one, and not foo-SNAPSHOT.jar)

Scratch that, we actually don't care if we scope the copied file to a given request as you suggested...

jvalkeal:
Well mark wanted to keep this easy for now(got that impression), thus reason for a 'naive caching'. Proper way would be to build a complete local cache repository which pretty much build unique key out from a 'unique' location string which brings more challenge as has from location is never really unique and some characters cannot be stored into a file name. Also for a perfect world we'd need to have a SHA to verify file but then whole resource abstraction infra need to be changed, etc. We should also handle parallel downloads and locking(takari from #11 has a locking impl), etc.

sabbyanandan:
I'd still vote for "naive caching" as interim solution until we design it the right way. This is blocking roadshow deliverables.

trisberg:
Can't they use Maven coordinates for the roadshow? It's pretty easy to set up a simple Maven repo on CF using the staticfile-buildpack. I'll see if I can put together some instructions tomorrow morning.

sabbyanandan:
The roadshow is entirely on pcfdev. Not quiet sure about `staticfile-buildpack` on the sandbox environment. Perhaps we could still provision on some other PCF instance and use it as remote repository.

markfisher:
rather than adding a UUID on our side, we should maybe just recommend that IF pointing to a snapshot version, then provide a unique version (e.g. the timestamp in the path)

jvalkeal:
maven locations are not cached as those resolves to local file automatically. if you point to http://repo.spring.io for snapshots you anyway need to use the real timestamped URL.

jvalkeal:
And besides saying "we should not aggressively cache too early" is totally invalid because maven resource will cache files every single time. What comes for a problem of having different files with a same name is something we need to resolve as I have a same issue with yarn when I copy file into hdfs.

markfisher:
I think we're ok punting on the potential for the remote resource to change (at least for now). If the resource is a Maven snapshot JAR (or something like it), then the URI should point to the actual *timestamped* file.

The other issue @ericbottard raised might be a bigger concern (collisions if the filename happens to be the same). Would a decent compromise for now be using the full URI instead of or along with filename somehow? Then, we might need to deal with escaping special characters.

WDYT @jvalkeal ? If that's not a trivial problem to solve either, then I'm inclined to push this as is for now with known limitations.

mbogoevici:
> maven locations are not cached as those resolves to local file automatically. if you point to http://repo.spring.io for snapshots you anyway need to use the real timestamped URL.

Of note, though, most Maven repositories will resolve the `-SNAPSHOT` in the URL to real artifacts, e.g. http://repo.spring.io/libs-snapshot-local/org/springframework/cloud/spring-cloud-dataflow-core/1.0.0.BUILD-SNAPSHOT/spring-cloud-dataflow-core-1.0.0.BUILD-SNAPSHOT.jar will resolve to latest snapshot.

markfisher:
right, but what we are saying is IF using plain old `http://` instead of `maven://` the recommendation would be to point to an explicit versioned JAR instead so you don't counter-intuitively maintain a stale snapshot in your cache

jvalkeal:
@markfisher only short term fix in this PR might be a one you mentioned(trying to come up with more unique file name/path from a location). I think anything beyond that for now would be worthless as that would mean to close this PR and make a bigger caching framework.

What about if we prefix that filename with a hash(calculated from location string). Hashing never produce unique results but combined with a filename, changes getting a collision is pretty slim?

jvalkeal:
Something like `scheme`-`location.hashCode()`-`resource.getFilename()`.

ericbottard:
What about hashing as in sha1 hashing instead ? Pretty simple still but way
less chances of collision.
Le 18 avr. 2016 9:57 AM, "Janne Valkealahti" <notifications@github.com> a
écrit :

jvalkeal:
Sha would be better but I'm not sure what crazy deps we need to pull in for that.

ericbottard:
None:

```
	public static void main(String[] args) throws Exception {
		String uri = "/some/path/foo.jar";
		MessageDigest md = MessageDigest.getInstance("SHA-1");
		System.out.println(formatAsHex(md.digest(uri.getBytes("UTF-8"))));
	}

	private static String formatAsHex(final byte[] bytes) {
		Formatter formatter = new Formatter();
		for (byte b : bytes) {
			formatter.format("%02x", b);
		}
		return formatter.toString();
	}

```

jvalkeal:
Rebased and force pushed my changes.

After adding that sha, logging look like:
```
module register --name foojee --type sink --uri http://repo.spring.io/libs-snapshot-local/org/springframework/cloud/stream/module/log-sink/1.0.0.BUILD-SNAPSHOT/log-sink-1.0.0.BUILD-20160418.094148-170.jar
module register --name foojee2 --type sink --uri http://repo.spring.io/libs-snapshot-local/org/springframework/cloud/stream/module/log-sink/1.0.0.BUILD-SNAPSHOT/log-sink-1.0.0.BUILD-20160415.193650-169.jar

module info --name sink:foojee

Caching file /tmp/deployer-resource-cache5302185025671673892/http-4d64a91bbf25db302f21479432bf131d20ddc56b-log-sink-1.0.0.BUILD-20160418.094148-170.jar as given location http://repo.spring.io/libs-snapshot-local/org/springframework/cloud/stream/module/log-sink/1.0.0.BUILD-SNAPSHOT/log-sink-1.0.0.BUILD-20160418.094148-170.jar

module info --name sink:foojee

Reusing cached file /tmp/deployer-resource-cache5302185025671673892/http-4d64a91bbf25db302f21479432bf131d20ddc56b-log-sink-1.0.0.BUILD-20160418.094148-170.jar as given location http://repo.spring.io/libs-snapshot-local/org/springframework/cloud/stream/module/log-sink/1.0.0.BUILD-SNAPSHOT/log-sink-1.0.0.BUILD-20160418.094148-170.jar

module info --name sink:foojee2

Caching file /tmp/deployer-resource-cache5302185025671673892/http-69131710c548aa033b850256f46fec33206ddedd-log-sink-1.0.0.BUILD-20160415.193650-169.jar as given location http://repo.spring.io/libs-snapshot-local/org/springframework/cloud/stream/module/log-sink/1.0.0.BUILD-SNAPSHOT/log-sink-1.0.0.BUILD-20160415.193650-169.jar

module info --name sink:foojee2

Reusing cached file /tmp/deployer-resource-cache5302185025671673892/http-69131710c548aa033b850256f46fec33206ddedd-log-sink-1.0.0.BUILD-20160415.193650-169.jar as given location http://repo.spring.io/libs-snapshot-local/org/springframework/cloud/stream/module/log-sink/1.0.0.BUILD-SNAPSHOT/log-sink-1.0.0.BUILD-20160415.193650-169.jar
```

spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/ShaUtils.java
```diff
@@ -0,0 +1,65 @@
...
+	 * Creates a sha1 out from a given data.
+	 *
+	 * @param data the data
+	 * @return the sha1 for data
+	 */
+	protected static String sha1(String data) {
+		try {
+			return new String(encodeHex(MessageDigest.getInstance("SHA-1").digest(data.getBytes("UTF-8"))));
+		}
+		catch (NoSuchAlgorithmException e) {
```
markfisher: I'd say these 2 exceptions should be `IllegalStateException` instead... that said, it's "guaranteed" to have SHA-1 according to Java docs

spring-cloud-deployer-resource-support/src/test/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoaderTests.java
```diff
@@ -16,24 +16,37 @@
...
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.junit.Rule;
 import org.junit.Test;
-
+import org.junit.rules.TemporaryFolder;
```
markfisher: should have an empty line before `org.springframework` imports

jvalkeal: we need to start adding sts settings to these projects if defaults are not good.

spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/ShaUtils.java
jvalkeal: I'd say you're right, don't know why IllegalArgumentException went there instead.

markfisher:
LGTM; merging

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-deployer/commit/ba12052f86c4d9484b5d30e9dfb263e64fbd68b6

Issue #25, opened 2016-03-25
LocalAppDeployer not using parent env


jvalkeal:
As deployer is using `ProcessLauncher` and then building environment for it, call
```
builder.environment().keySet().retainAll(ENV_VARS_TO_INHERIT);
```
is made to not inherit 'stuff' from boot. However this is causing i.e. system character set from a parent process not available which totally breaks encoding if non unicode characters are used.

This is a reason why in tests we got failures:
```
Expected value to be equal to '&'"|< ?\(', but was '&'"|< ??\('
```
where French character `é` was used.

We need to figure out how to inherit environment from OS while removing something what deployer may have added to environment.

pperalta:
https://github.com/spring-cloud/spring-cloud-deployer/pull/41

PR #41, opened 2016-04-12
Configurable regex for env var retention

pperalta:
Replaced hard coded set of env variables with a configurable string array of regular expressions for var retention.

The check for Windows to include `TMP` was removed for simplicity - is this required or can we include this on other platforms?

This fixes https://github.com/spring-cloud/spring-cloud-deployer/issues/25.


jvalkeal:
I'll look this closely tomorrow, but according to http://stackoverflow.com/questions/1924136/environment-variable-to-control-java-io-tmpdir, windows will get temp location from winapi call, while unix is default location is hardcoded. Yeah, this is fun.

jvalkeal:
lgtm, merging

PR #43, opened 2016-04-14
Adding a way to slow down redeployment test

trisberg:
- On Marathon, we need some time to wait for the platform to cleaan up before reusing the same app id. It's noot enough to wait for the app to not exists anymore.


ericbottard:
Rebased and merged as 8392845

PR #50, opened 2016-04-19
Guard getFile with all exceptions

### LINKED ISSUE #49: Guard getFile with all exceptions in DelegatingResourceLoader
We check if `Resource` exists as file by catching `IOException`. However it looks like we have implementations(like `HdfsResource`) which throw `UnsupportedOperationException` and error is then propagated. Should really catch `Exception` as if something is thrown, file doesn't exist.

jvalkeal:
- Add better catch in DelegatingResourceLoader
  for existsAsFile().
- Fixes #49

### DIFF
--- a/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
+++ b/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
@@ -159,7 +159,7 @@ public class DelegatingResourceLoader implements ResourceLoader {
 			resource.getFile();
 			return true;
 		}
-		catch (IOException e) {
+		catch (Exception e) {
 		}
 		return false;
 	}


spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
```diff
@@ -159,7 +159,7 @@ private static boolean existsAsFile(Resource resource) {
 			resource.getFile();
 			return true;
 		}
-		catch (IOException e) {
+		catch (Exception e) {
```
markfisher: should we log something, even if it's at *debug* level?

markfisher: maybe not since this is only called internally anyways... implementation detail

will merge

PR #51, opened 2016-04-22
add overwrite boolean param to registry populator

markfisher:
required by https://github.com/spring-cloud/spring-cloud-dataflow/pull/549

### DIFF
--- a/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/registry/UriRegistryPopulator.java
+++ b/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/registry/UriRegistryPopulator.java
@@ -40,6 +40,7 @@ import org.springframework.util.Assert;
  * provided by {@link #resourceLoader}.
  *
  * @author Patrick Peralta
+ * @author Ilayaperumal Gopinathan
  */
 public class UriRegistryPopulator implements ResourceLoaderAware {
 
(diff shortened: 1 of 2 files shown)


spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/registry/UriRegistryPopulator.java
```diff
@@ -73,6 +72,15 @@ public void setResourceLoader(ResourceLoader resourceLoader) {
 				for (String key : properties.stringPropertyNames()) {
 					try {
 						URI uri = new URI(properties.getProperty(key));
+						if (!overwrite) {
+							try  {
+								registry.find(key);
+							}
+							catch (IllegalArgumentException e) {
+								// this key already exists
```
ilayaperumalg: It looks like the `IllegalArgumentException` is thrown only when the key is not found. Hence, I think we need to register in that case instead of `continue` here and if the key is found then we can continue.

ilayaperumalg: @markfisher Pushed some changes to my branch [here] (https://github.com/ilayaperumalg/spring-cloud-deployer/commit/2ea3d30d8d3b546c6ab250f97baac0f9766ee015) for the above comment with a test that showcases it. If that looks good to you, I can merge this PR along with those changes.

markfisher: thanks @ilayaperumalg - good catch! your changes look good except one minor detail: *exist* (no s at the end) in the catch block comment

markfisher:
I've incorporated your commit @ilayaperumalg. Thanks!

PR #53, opened 2016-04-25
Updating Docker image

trisberg:
- changing to use new official image java:8-alpine

- changing to use springcloud account instead of springcloudstream since this isn't stream related


jvalkeal:
lgtm, merging

PR #54, opened 2016-04-26
Make the test logger protected for classes extending this

trisberg:


spring-cloud-deployer-spi-test/src/main/java/org/springframework/cloud/deployer/spi/test/AbstractAppDeployerIntegrationTests.java
```diff
@@ -73,7 +73,7 @@
 @RunWith(SpringJUnit4ClassRunner.class)
 public abstract class AbstractAppDeployerIntegrationTests {
 
-	private static final Logger log = LoggerFactory.getLogger(AbstractAppDeployerIntegrationTests.class);
+	protected static final Logger log = LoggerFactory.getLogger(AbstractAppDeployerIntegrationTests.class);
```
markfisher: could make this non-static and call `this.getClass()` instead

markfisher: but it should indeed be `protected` so it can be used within subclasses

markfisher:
merged, but also made `log` non-static so it could be created with the concrete class: https://github.com/spring-cloud/spring-cloud-deployer/commit/801d925a6611e8cacaf1d1a3ccdf01a529228ab2

PR #56, opened 2016-05-03
LocalTaskLauncher not using parent env

### LINKED ISSUE #42: LocalTaskLauncher not using parent env
Work done in #25 should also done for tasks. We're starting to have a lot of duplication so it's best to wait #20 which adds a share parent class for common functionality.

jvalkeal:
- Refactor LocalTaskLauncher to use same
  ProcessBuilder concepts than LocalAppDeployer.
- Moved some common functionality into a parent
  AbstractDeployerSupport.
- Fixes #42


cppwfs:
LGTM

PR #58, opened 2016-05-04
Improved local application shutdown

pperalta:
Introduced timeout for application shutdown via `/shutdown` endpoint
Reset flag on `InterruptedException`
Made shutdown timeout configurable

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
@@ -43,6 +43,7 @@ import org.springframework.cloud.deployer.spi.app.AppStatus;
 import org.springframework.cloud.deployer.spi.app.DeploymentState;
 import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
 import org.springframework.core.io.Resource;
+import org.springframework.http.ResponseEntity;
 import org.springframework.util.SocketUtils;
 import org.springframework.util.StringUtils;
 import org.springframework.web.client.ResourceAccessException;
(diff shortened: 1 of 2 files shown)


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
```diff
@@ -206,13 +207,40 @@ public AppStatus status(String id) {
...
-			restTemplate.postForObject(instance.url + "/shutdown", null, String.class);
-			instance.process.waitFor();
+			int timeout = getLocalDeployerProperties().getShutdownTimeout();
+			if (timeout > 0) {
+				ResponseEntity<String> response = restTemplate.postForEntity(
+						instance.url + "/shutdown", null, String.class);
+				if (response.getStatusCode().is2xxSuccessful()) {
+					long timeoutTimestamp = System.currentTimeMillis() + timeout;
+					while (isAlive(instance.process) && System.currentTimeMillis() < timeoutTimestamp) {
+						Thread.sleep(1000);
```
markfisher: If we're going to wait a second here anyways, I think it might be ok to have the timeout granularity (the `TimeUnit` enum value basically) be *seconds* rather than milliseconds. Otherwise, any value would be effectively rounded to the next second anyways. WDYT?

pperalta: Yes, very good point.

markfisher: I can handle it on merge... I think keeping the property name `shutdownTimeout` is fine as long as it's documented to be in SECONDS.

markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-deployer/commit/d9a2f0e7a9dc6e644c8a6f00d09e870657096cba
along with: https://github.com/spring-cloud/spring-cloud-deployer/commit/7596e49f029eb418d3a19e5128cfc7bb93aed5fb

PR #59, opened 2016-05-05
Classpath based application launching

pperalta:
Introduced option to supply a main class and classpath instead of an executable jar. This is useful for situations such as testing where a boot app isn't in an executable jar.

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractDeployerSupport.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractDeployerSupport.java
@@ -17,6 +17,7 @@
 package org.springframework.cloud.deployer.spi.local;
 
 import java.util.ArrayList;
+import java.util.Map;
 
 import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
 import org.springframework.util.Assert;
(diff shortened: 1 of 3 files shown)


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractDeployerSupport.java
```diff
@@ -60,8 +61,17 @@ final protected LocalDeployerProperties getLocalDeployerProperties() {
...
+		Map<String, String> envProps = request.getEnvironmentProperties();
+		if (envProps.containsKey("main") && envProps.containsKey("classpath")) {
+			commands.add("-cp");
+			commands.add(envProps.get("classpath"));
+			commands.add(envProps.get("main"));
+		}
+		else {
+			commands.add("-jar");
+			commands.add(jarPath);
+		}
```
markfisher: Should we handle the case where `classpath` or `main` is provided without the other? (currently would end up going the `-jar` route)... and on a related note, it seems odd to be handling this decision while already in a method where the `jarPath` String has been passed in as a parameter. I wonder if we should be somehow checking for this main-class alternative somewhere earlier?

pperalta: > Should we handle the case where classpath or main is provided without the other?

How about an `IllegalStateException`?

>  it seems odd to be handling this decision while already in a method where the jarPath String has been passed in as a parameter

Agreed - it should be renamed to `buildExecutionCommand`

> I wonder if we should be somehow checking for this main-class alternative somewhere earlier?

I'm OK with it being here, but if you have any suggestions I'm happy to discuss.

markfisher: I think I have an idea for refactoring this a bit, but will try that after merging this PR.

markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-deployer/commit/94b5fcc2b6d82cf00a2ba6622343b401176ec011

other changes to address comments above are forthcoming

PR #72, opened 2016-06-09
Verify deployment state during deploy and undeploy exited instances

### LINKED ISSUE #71: Handle partial deployment in local app deployer
When the stream is partially deployed, subsequent `destroy` and re-deploy results in issue mentioned in: https://github.com/spring-cloud/spring-cloud-dataflow/issues/659

ilayaperumalg:
- In local deployer the deployment status should be checked before deploying the app
  - This change removes the existing check to verify the running processes which could be wrong when the app exited but the map `running` still has references of exited instances

 - When checking the `status`, undeploy (cleanup the `running` map with exited instances) can be invoked

This resolves #71


markfisher:
LGTM; merged via https://github.com/spring-cloud/spring-cloud-deployer/commit/0a8a8415569b98d7d6d94e814bcc49cd16d7c1ed

PR #76, opened 2016-06-21
Add ability to specify authentication for remote repositories.

### LINKED ISSUE #75: spring-cloud-deployer-resource-maven should support remote repository authentication
Following on from [this Gitter conversation](https://gitter.im/spring-cloud/spring-cloud?at=57691e73bd67400679dba6ce), it should be possible to provide authentication details for remote Maven repositories the same way authentication is [supported for HTTP proxies](https://github.com/spring-cloud/spring-cloud-deployer/blob/master/spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenArtifactResolver.java#L98).

donovanmuller:
Fixes #75

### DIFF
--- a/spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenArtifactResolver.java
+++ b/spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenArtifactResolver.java
@@ -63,6 +63,7 @@ import org.springframework.util.StringUtils;
  * @author Mark Fisher
  * @author Marius Bogoevici
  * @author Ilayaperumal Gopinathan
+ * @author Donovan Muller
  */
 class MavenArtifactResolver {
 
@@ -98,19 +99,7 @@ class MavenArtifactResolver {
 		if (isProxyEnabled() && proxyHasCredentials()) {
 			final String username = this.properties.getProxy().getAuth().getUsername();
 			final String password = this.properties.getProxy().getAuth().getPassword();
-			this.authentication = new Authentication() {
-				@Override
-				public void fill(AuthenticationContext context, String key, Map<String, String> data) {
-					context.put(AuthenticationContext.USERNAME, username);
-					context.put(AuthenticationContext.PASSWORD, password);
-				}
-
-				@Override
-				public void digest(AuthenticationDigest digest) {
-					digest.update(AuthenticationContext.USERNAME, username,
-							AuthenticationContext.PASSWORD, password);
-				}
-			};
+			this.authentication = newAuthentication(username, password);
 		}
 		else {
 			this.authentication = null;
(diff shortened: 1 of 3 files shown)


donovanmuller:
This PR allows you to pass in credentials for a remote Maven repository that requires authentication.
So the following is now possible:

```
$ java -jar spring-cloud-dataflow-server-kubernetes-1.0.0.BUILD-SNAPSHOT.jar \
  --maven.remoteRepositories[0].url=http://my.repo.com \
  --maven.remoteRepositories[0].auth.username=user \
  --maven.remoteRepositories[0].auth.password=user123
```

The `maven.remoteRepositories[0].` is a little ugly so would be interested if there is a better way

donovanmuller:
Thinking about it, perhaps the current way of declaring remote repositories should be preserved as well as supporting the capability to add auth credentials?

I.e. support the current:

```
$ java -jar spring-cloud-dataflow-server-kubernetes-1.0.0.BUILD-SNAPSHOT.jar \
    --maven.remoteRepositories=http://my.repo1.com,http://my.repo2.com
```

as well as supporting the ability to add auth credentials [as above](https://github.com/spring-cloud/spring-cloud-deployer/pull/76#issuecomment-227435172)

Thoughts?

ilayaperumalg:
Hi @donovanmuller, thanks for the PR. I have made some changes on top of your PR to allow the `remoteRepository` properties passed using an alias name. 

Please see [here] (https://github.com/ilayaperumalg/spring-cloud-deployer/commit/e6c45d94a114a516c9cb600466e0b6d481b7ca15)

For instance,

```
java -jar spring-cloud-dataflow-server-local-1.0.0.BUILD-SNAPSHOT.jar  \
--maven.localRepository=mylocal --maven.remote-repositories.test1.url=remoteURL1 \
--maven.remote-repositories.test1.auth.username=user1 \
--maven.remote-repositories.test1.auth.password=pass1 --maven.proxy.host=proxy1 \
--maven.proxy.port=9010 --maven.proxy.auth.username=proxyuser1 \
--maven.proxy.auth.password=proxypass1

```
By default, the spring repo URL is always updated along with any other remote repositories passed in. 

I believe this way of specifying the properties would be better. Let me know how this works in your environment (especially with the auth for the remote repositories).

Thanks!

donovanmuller:
@ilayaperumalg much cleaner 😄. Will [give it](https://github.com/ilayaperumalg/spring-cloud-deployer/commit/e6c45d94a114a516c9cb600466e0b6d481b7ca15) a try now...

ilayaperumalg:
@donovanmuller thank you! There is one limitation with this approach. Since we use `Map` of properties that are as well `nested`, we won't be able to set the `maven` properties via **environment** properties. Hence, the most preferable way would be set via `cmdline` properties when starting the server.

In a way, this holds true for your initial approach as well (the environment properties can't be set with identifiers such as `[]` )

donovanmuller:
@ilayaperumalg I can confirm your [changes](https://github.com/ilayaperumalg/spring-cloud-deployer/commit/e6c45d94a114a516c9cb600466e0b6d481b7ca15) work.

As for the environment properties, that is not an issue for me, not sure about others?

ilayaperumalg:
thanks for the confirmation. For the environment properties issue, it can be worked around using the environment property `SPRING_APPLICATION_JSON` if those required to be set as environment properties only. The changes and along with this work around for environment properties need to be documented once confirmed. @markfisher what do you think?

markfisher:
I like the use of aliases. As far as I understand, that does not have any limitations that are not also present when using indices, yet it is clearer and more explicit (and therefore less error-prone). If the only limitation is direct use of the nested maps as env vars, the only alternative would be `SPRING_APPLICATION_JSON` anyways - as you suggested. That being an option as well as the ability to specify all of these properties in a YAML file (which I would expect to be the preferred option in fact), seems worth the tradeoff of being able to configure these properties clearly, for multiple remote repositories.

ilayaperumalg:
LGTM, rebased and merged the changes 7d10d525a1d16ceccae881fae6c44a3d0baa40ac along with the changes to handle remote-repositories configuration properties e6c45d94a114a516c9cb600466e0b6d481b7ca15.

@donovanmuller thank you again for the PR. Looking forward to more contributions!

PR #81, opened 2016-06-29
Add ability to override default remote repo

ilayaperumalg:
- When user provides explicit remote repository, then the default remote repo (maven central) should be removed from the list of remote repositories

### DIFF
--- a/spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenProperties.java
+++ b/spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenProperties.java
@@ -17,10 +17,8 @@
 package org.springframework.cloud.deployer.resource.maven;
 
 import java.io.File;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Random;
 
 /**
  * Configuration Properties for Maven.
@@ -43,13 +41,10 @@ public class MavenProperties {
 	 */
 	private String localRepository = DEFAULT_LOCAL_REPO;
 
-	private static final String DEFAULT_REMOTE_REPO_KEY = "default" + new Random().nextInt();
-
 	/**
 	 * Locations of remote maven repositories from which artifacts will be downloaded, if not available locally.
 	 */
-	private Map<String, RemoteRepository> remoteRepositories = new HashMap<>(
-			Collections.singletonMap(DEFAULT_REMOTE_REPO_KEY, new RemoteRepository("http://repo1.maven.org/maven2")));
+	private Map<String, RemoteRepository> remoteRepositories = new HashMap<>();
 
 	/**
 	 * Whether the resolver should operate in offline mode.
@@ -74,10 +69,6 @@ public class MavenProperties {
 	private Integer requestTimeout;
 
 	public Map<String, RemoteRepository> getRemoteRepositories() {
-		// Since the remote repositories is a map, we need to remove the default if user provides explicit repo.
-		if (remoteRepositories.size() > 1 && remoteRepositories.containsKey(DEFAULT_REMOTE_REPO_KEY)) {
-			this.remoteRepositories.remove(DEFAULT_REMOTE_REPO_KEY);
-		}
 		return remoteRepositories;
 	}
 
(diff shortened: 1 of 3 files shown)


spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenProperties.java
```diff
@@ -41,10 +43,13 @@
 	 */
 	private String localRepository = DEFAULT_LOCAL_REPO;
 
+	private static final String DEFAULT_REMOTE_REPO_KEY = "default" + new Random().nextInt();
```
ilayaperumalg: Using `random` value here in order not to collide with any other user provided alias with the key `default`.

spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenProperties.java
```diff
@@ -41,10 +43,13 @@
...
+
 	/**
 	 * Locations of remote maven repositories from which artifacts will be downloaded, if not available locally.
 	 */
-	private Map<String, RemoteRepository> remoteRepositories = new HashMap<>();
+	private Map<String, RemoteRepository> remoteRepositories = new HashMap<>(
+			Collections.singletonMap(DEFAULT_REMOTE_REPO_KEY, new RemoteRepository("http://repo1.maven.org/maven2")));
```
ilayaperumalg: Since having direct singleton map here makes it `immutable`, creating a new `HashMap`.

spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenProperties.java
```diff
@@ -69,6 +74,10 @@
 	private Integer requestTimeout;
 
 	public Map<String, RemoteRepository> getRemoteRepositories() {
+		// Since the remote repositories is a map, we need to remove the default if user provides explicit repo.
+		if (remoteRepositories.size() > 1 && remoteRepositories.containsKey(DEFAULT_REMOTE_REPO_KEY)) {
```
ilayaperumalg: Since `remoteRepositories`

markfisher: why is all this necessary? If the user provides explicit repos, they overwrite the whole map, right?

spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenProperties.java
markfisher: why do we need a random key here? I would expect just using "mavenCentral" as the key on line 52

ilayaperumalg: ok, `mavenCentral` looks like a better alternative. will use that here.

spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenProperties.java
ilayaperumalg: yeah, I was hoping to get the map replacing the existing on via the `setter`. But, it turns out that the map just gets `updated` instead of `replaced`. Hence, I went ahead with the above approach. What do you think?

markfisher:
this is what I was thinking could be simpler: https://github.com/markfisher/spring-cloud-deployer/commit/b15661ce87d6a4bfcc7c8fe92dd6880da89bb16f

tests fail, but that's because I think they need to have the config property set

markfisher:
the build is failing for me locally

ilayaperumalg:
It ran fine for me though. what error do you see?

markfisher:
actually I believe it's because of a chicken-and-egg situation with the test app being `1.0.1.BUILD-SNAPSHOT` itself

markfisher:
so... works ok with `mvn clean install` (just not with `mvn clean package` - at least not until we get one build from the CI server so the remote snapshot is available)

markfisher:
LGTM; squashed and merged as https://github.com/spring-cloud/spring-cloud-deployer/commit/ff7d7e3b71f4460f0de443d065787bfb643d9b0f

PR #84, opened 2016-07-01
do not pass deployment props to the app

### LINKED ISSUE #83: Local deployer should not copy deployment properties to the environment.
https://github.com/spring-cloud/spring-cloud-deployer/blob/master/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java#L101

Should be removed.

markfisher:
fixes #83


ilayaperumalg:
LGTM, merged along with the changes:

Fix LocalTaskLauncher

      - not passing deployment props to app args

    Also remove `group-deployment-id` usage as it is not a standard parameter and only used in local app deployer

PR #86, opened 2016-07-01
update doc for app properties

markfisher:


jvalkeal:
Think I'm fine with those changes.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/core/AppDefinition.java
```diff
@@ -29,7 +29,7 @@
  * A deployer may not modify properties in this class as those are meant for an
  * actual app as is. The deployer's only responsibility is to pass those
  * properties into a runtime environment in a way that they are available in a
- * running application.
+ * running application. For example, they would typically be passed as env vars.
```
markpollack: ATM, only 1 of the 5 impls does this, mesos.

trisberg: I would leave this out, we don't know what is typical yet ...

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/core/AppDeploymentRequest.java
```diff
@@ -34,11 +34,12 @@
  * app is executed, such as allowed memory, cpu or simply a way to define
  * colocation like node labeling.
  *
- * For passing properties or parameters into the app itself, use
- * {@link AppDefinition#getProperties()}.
+ * For passing properties into the app itself, use {@link AppDefinition#getProperties()}.
+ * Those will typically be passed as env vars, unless an alternative approach is better
+ * suited to the target platform. Each deployer implementation should clearly document
+ * how it handles these properties.
  *
```
trisberg: I would say they could be passed as env vars or as --key=value style command line args

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/core/AppDefinition.java
markfisher: ok or I can change to:

```
For example, they could be passes as env vars.
```

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/core/AppDeploymentRequest.java
markfisher: will update

sobychacko:
merged.

PR #92, opened 2016-07-07
Add instance index key as constant in AppDeployer

### LINKED ISSUE #91: Add constant for "INSTANCE_INDEX" to AppDeployer
Any implementation that supports partitioning via the count/indexed combo need to set an env var of "INSTANCE_INDEX", we should add a constant to the AppDeployer interface for this purpose.

ilayaperumalg:
This resolves #91

### DIFF
--- a/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppDeployer.java
+++ b/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppDeployer.java
@@ -56,6 +56,12 @@ public interface AppDeployer {
 	static final String INDEXED_PROPERTY_KEY = PREFIX + "indexed";
 
 	/**
+	 * Constant property key to be used in the deployer implementation as an environment variable
+	 * to set the app instance index (if needed).
+	 */
+	static final String INSTANCE_INDEX_KEY = "INSTANCE_INDEX";
+
+	/**
 	 * Deploy an app using an {@link AppDeploymentRequest}. The returned id is
 	 * later used with {@link #undeploy(String)} or {@link #status(String)} to
 	 * undeploy an app or check its status, respectively.


spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppDeployer.java
```diff
@@ -56,6 +56,12 @@
 	static final String INDEXED_PROPERTY_KEY = PREFIX + "indexed";
 
 	/**
+	 * Constant property key to be used in the deployer implementation as an environment variable
```
ericbottard: I wouldn't say "Constant property key" here, because it is confusing given that other constants are named `*_PROPERTY_KEY` (and play a very different role).
I'll change this part on merge

ericbottard: Actually, on second thought, this *is* a property key, as this technically hasn't to be an env_var.
I'd consider changing the javadoc to something like this, wdyt:
```java
	/**
	 * The property to be set at each instance level to specify the sequence number
	 * amongst 0 to N-1, where N is the value of the {@value #COUNT_PROPERTY_KEY} property.
	 * Specified as CAPITAL_WITH_UNDERSCORES as this is typically passed as an environment
	 * variable, but when targeting a Spring app, other variations may apply.
	 *
	 * @see #INDEXED_PROPERTY_KEY
	 */
	static final String INSTANCE_INDEX_PROPERTY_KEY = "INSTANCE_INDEX";
```

ericbottard: Note that I'm not saying "deployment property", because this is to be set at each app level *by* the deployer, as opposed to deployment properties that are *read* by the deployer

ilayaperumalg: Sure, I like it.

ericbottard:
Merged as 40c3e33

PR #93, opened 2016-07-07
Added default constructor.

cppwfs:
Since we now support "empty" or null loaders, a user should be able to use a default constructor.


spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
```diff
@@ -68,6 +68,13 @@
 
 	/**
 	 * Instantiates a new delegating resource loader.
+	 */
+	public DelegatingResourceLoader() {
```
markfisher: now this makes me wonder if the name is wrong... since clearly it does not *need* to delegate. There are only 2 reasons to use this class without a map of delegates:

1. to take advantage of its file-caching behavior
2. to rely on this specific type being present within the application context

cc @jvalkeal

Issue #95, opened 2016-07-11
Enhance TCK wrt count, instanceIndex, and deploymentProperties


ericbottard:
Now that the role of `deploymentProperties` has been clarified, the abstract test class that is used for the TCK could
* validate that `count` is properly handled (esp. given that the in-process option is no longer available)
* validate that `deploymentProperties` are not passed to the deployed app

trisberg:
What about "GROUP_PROPERTY_KEY", don't  think we have any tests verifying that it is used, although, the intended expectations of its use isn't really specified by the SPI.

trisberg:
Should we make a 1.0.x branch and switch to 1.1.0.BUILD-SNAPSHOT or does this belong in 1.0.3.BUILD-SNAPSHOT?

sabbyanandan:
This is not big enough to switch to 1.1.x, yet. @markfisher: thoughts?

trisberg:
it's labeled as 1.1.0.M1

markfisher:
I think we went back and forth a few times. At one point, the issue was labeled 1.1.0.M1. The question is whether that was our final decision. I don't recall. CC @markpollack

ericbottard:
Given that it adds tests only, and that local deployer passes as is, i'd
say maintenance is fine. (ie no bug fixed, let alone no new feature).

Le 22 juil. 2016 10:51 PM, "Thomas Risberg" <notifications@github.com> a
écrit :

markfisher:
I'm okay with that, and it is the simpler option ;)

markpollack:
+1

PR #97, opened 2016-07-18
Enhance TCK with additional tests

ericbottard:
Fixes spring-cloud/spring-cloud-deployer#95


spring-cloud-deployer-spi-test/src/main/java/org/springframework/cloud/deployer/spi/test/AbstractAppDeployerIntegrationTests.java
```diff
@@ -251,10 +254,84 @@ public void testParameterPassing() {
...
+	 * Tests that command line arguments (ordered strings) can be passed to a deployed app,
+	 * including values that typically require special handling.
+	 */
+	@Test
+	public void testCommandLineArgumentsPassing() {
+		Map<String, String> properties = new HashMap<>();
+		AppDefinition definition = new AppDefinition(randomName(), properties);
+		Map<String, String> deploymentProperties = new HashMap<>();
+
+		List<String> cmdLineArgs = Arrays.asList("--commandLineArgValueThatMayNeedEscaping=d" + DeployerIntegrationTestProperties.FUNNY_CHARACTERS);
```
trisberg: looks like the extra 'd' character makes the tests fail, will remove before merging unless there was a specific reason for it to be there

ericbottard: Hmmm, this is not good. I added it to test that it indeed would fail, but must have forgotten to remove it. At first I was surprised that an arg that contained a space could be passed as is, so doubted of the quality of the test. I then must have been carried away.

PR #100, opened 2016-08-16
Fix LocalDeployer to fail faster in status()

ericbottard:
Also, have a better default name for deployed apps in tests


markfisher:
LGTM; merged

PR #106, opened 2016-08-26
Add infrastructure for Tasks TCK

ericbottard:
Other tests to be added later

Fixes spring-cloud/spring-cloud-deployer#104

### DIFF
--- a/spring-cloud-deployer-spi-test/src/main/java/org/springframework/cloud/deployer/spi/test/AbstractTaskLauncherIntegrationTests.java
+++ b/spring-cloud-deployer-spi-test/src/main/java/org/springframework/cloud/deployer/spi/test/AbstractTaskLauncherIntegrationTests.java
@@ -22,10 +22,9 @@ import static org.junit.Assert.assertThat;
 import org.hamcrest.BaseMatcher;
 import org.hamcrest.Description;
 import org.hamcrest.Matcher;
+import org.hamcrest.Matchers;
 import org.junit.Test;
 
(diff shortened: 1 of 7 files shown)


spring-cloud-deployer-spi-test/src/main/java/org/springframework/cloud/deployer/spi/test/AbstractTaskLauncherIntegrationTests.java
```diff
@@ -0,0 +1,129 @@
...
+
+
+
+
+	/**
+	 * Return a resource corresponding to the spring-cloud-deployer-spi-test-app app suitable for the target runtime.
+	 *
+	 * The default implementation returns an uber-jar fetched via Maven. Subclasses may override.
+	 */
+	protected Resource integrationTestProcessor() {
```
ericbottard: That will need to be replaced by a task-specific equivalent.

Also, ideally the app-specific artifact should change name (as would this method), but this is a bit disruptive.
Hopefully `@Override` helps here..

mminella:
Was the branch for this PR deleted?  I can't seem to find it either when I fetch from @ericbottard repo or when I browse it via Github...

ericbottard:
You scared me @mminella . It's here: https://github.com/ericbottard/spring-cloud-deployer/tree/issue_104_TCK_tasks

mminella:
LGTM.  Merged as 7294950f3470132a7d23fcf31c49477bc47065ae 

P.S.  Sorry about the scare.  I was looking in the cf deployer project :(

PR #113, opened 2016-09-27
Fix local deployer log config property

### LINKED ISSUE #112: Local deployer log directory config property doesn't work
Currently, the property `deployer.local.working-directories-root` doesn't seem to work when set. When set with a String value, following exception occurs:
```
codes [typeMismatch.deployer.local.workingDirectoriesRoot,typeMismatch.workingDirectoriesRoot,typeMismatch.java.nio.file.Path,typeMismatch]; arguments [org.springframework.context.support.DefaultMessageSourceResolvable: codes [deployer.local.workingDirectoriesRoot,workingDirectoriesRoot]; arguments []; default message [workingDirectoriesRoot]]; default message [Failed to convert property value of type [java.lang.String] to required type [java.nio.file.Path] for property 'workingDirectoriesRoot'; nested exception is java.lang.IllegalStateException: Cannot convert value of type [java.lang.String] to required type [java.nio.file.Path] for property 'workingDirectoriesRoot': no matching editors or conversion strategy found]
```

ilayaperumalg:
Resolves #112


markfisher:
LGTM; merged

Issue #114, opened 2016-09-27
Upgrade to Camden


sabbyanandan:
As a developer, I'd like to upgrade to Camden, so all the upstream release dependencies are synchronized.

ericbottard:
Reopening, as full upgrade should also have upgraded the spring-cloud-build dependency, which brings in its share of required changes.

Will issue a second PR, that also bumps master to 1.1.0.BUILD-SNAPSHOT

Issue #119, opened 2016-10-05
Need new branch for 1.0.x and master to update to `1.1.0`


ilayaperumalg:

ericbottard:
Done

PR #122, opened 2016-10-12
Upgrade to Spring Cloud Camden

ericbottard:
Fixes https://github.com/spring-cloud/spring-cloud-deployer/issues/114

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,7 +21,7 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<java.version>1.7</java.version>
 		<spring-framework.version>4.3.3.RELEASE</spring-framework.version>
-		<spring-cloud.version>Brixton.SR6</spring-cloud.version>
+		<spring-cloud.version>Camden.RELEASE</spring-cloud.version>
 		<aether.version>1.0.2.v20150114</aether.version>
 		<maven.version>3.1.0</maven.version>
 	</properties>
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
LGTM, merging.

trisberg:
Shouldn't we have made this change for 1.1.0.BUILD-SNAPSHOT? We are still on a 1.0.4.BUILD-SNAPSHOT version. That's what we have done for SCDF, 1.0.x uses Brixton while 1.1.x uses Camden.

ericbottard:
Ah, good point. Now, corollary question: should we move to 1.1.0.BUILD-SNAPSHOT? :)

trisberg:
I would say we should, after we create a 1.0.x branch on the commit preceding this change - relates to #119

trisberg:
Why, any later changes including https://github.com/spring-cloud/spring-cloud-deployer/commit/b2e0a2eb8736fdbd6b07b67fee1b977aaef53093 would be a fit for 1.0.x, especially the Brixton upgrade

ericbottard:
Ah, correct. Did not realize there were so few commits after latest release and they were appropriate

ericbottard:
I'm on it. Will push a branch, create a new CI plan and update master to be 1.1.0.B-S

trisberg:
We should also update the parent (spring-cloud-build) to use the latest 1.2.x to match Camden

PR #127, opened 2016-10-13
Create new Abstract TestRule class

### LINKED ISSUE #126: Create new Abstract TestRule class
We need to replace the use of `AbstractExternalResourceTestSupport` from the `spring-cloud-stream-test-support-internal` project since that is a Stream internal class and the Stream BOM no longer includes the version for this artifact

trisberg:
- ported from the Spring Cloud Stream project's internal support class previously used for the deployer implementations tests

Resolves #126


ericbottard:
Merged as 1b602c1 after having changed the prefix used

Issue #128, opened 2016-10-17
Add general purpose commandLine parsing utility


ericbottard:
Useful when dealing with a configuration property that is meant to be passed to a spawned process, parsing of command line args (dealing with single and double quotes, etc) should be available in core deployer and re-used in specific implementations.

see https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/pull/64

PR #129, opened 2016-10-17
Add a simple command line tokenizer

ericbottard:
Fixes https://github.com/spring-cloud/spring-cloud-deployer/issues/128


ericbottard:
> should the junit dependency be test scope?

Ooops, yes (if you intended to make this a comment on the junit dependency (not the tokenizer itself)

PR #135, opened 2016-10-21
Fix file name for non-File resources

### LINKED ISSUE #96: Allow getFilename() to contain "/"s
As a user, I'm trying to resolve maven artifacts from an S3 bucket; however, I'm running into an exception because the target cache folder is not created if the file name includes `/` character.

ilayaperumalg:
- If the resource loader gets the resource whose filename doesn't exactly return the last part of the filepath and also if the filename happens to have special characters for some reason, then the cache filename for that resource will remove the invalid/not supported characters
  - Only word characters, numbers, _ and - are allowed
  - Add tests

Fixes #96

### DIFF
--- a/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
+++ b/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
@@ -51,6 +51,7 @@ import org.springframework.util.FileCopyUtils;
  *
  * @author Mark Fisher
  * @author Janne Valkealahti
+ * @author Ilayaperumal Gopinathan
  */
 public class DelegatingResourceLoader implements ResourceLoader, ResourceLoaderAware {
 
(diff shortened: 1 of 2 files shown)


spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
```diff
@@ -117,7 +118,9 @@ public Resource getResource(String location) {
 				return resource;
 			}
 			else {
-				String cacheName = scheme + "-" + ShaUtils.sha1(location) + "-" + resource.getFilename();
+				String fileName = resource.getFilename();
```
ilayaperumalg: Yeah, good idea. Given we display the cache file name in the console, I think this is good.

ericbottard: Frankly, having had a second look at the code, why do we even bother with the filename.
The `location` already includes the filename (in some way or the other) and is the perfect cache key.

`sha1(scheme + location)` or `scheme + sha1(location)` is the simplest and more sensible solution.

I can change on merge

ericbottard:
Edited and merged as 36bdecc

PR #137, opened 2016-10-24
Fixing the way we handle exit code and "unknown" state

### LINKED ISSUE #136: Local deployer should report stream apps with any exit code as failed
Local deployer reports failed apps as "undeployed" if exit code is 0. This is the case when an app can't connect to zookeeper/kafka.

This causes streams to be reported as "undeployed" instead of "failed". See https://github.com/spring-cloud/spring-cloud-dataflow/issues/960 for more detail.

trisberg:
- apps can fail and return 0 exit code, we should treat any exited stream apps as failed
- a call to status should not change the state of the system, removing code that undeploys "unknown" app

Resolves #136

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
@@ -159,10 +159,6 @@ public class LocalAppDeployer extends AbstractLocalDeployerSupport implements Ap
 			}
 		}
 		AppStatus status = builder.build();
-		// Make sure to explicitly undeploy any exited instances that have status undeployed.
-		if (DeploymentState.undeployed.equals(status.getState())) {
-			undeploy(id);
-		}
 		return status;
 	}
 
@@ -229,12 +225,7 @@ public class LocalAppDeployer extends AbstractLocalDeployerSupport implements Ap
 			Integer exit = getProcessExitValue(process);
 			// TODO: consider using exit code mapper concept from batch
 			if (exit != null) {
-				if (exit == 0) {
-					return DeploymentState.undeployed;
-				}
-				else {
-					return DeploymentState.failed;
-				}
+				return DeploymentState.failed;
 			}
 			try {
 				HttpURLConnection urlConnection = (HttpURLConnection) baseUrl.openConnection();


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
```diff
@@ -229,12 +225,7 @@ public DeploymentState getState() {
 			Integer exit = getProcessExitValue(process);
 			// TODO: consider using exit code mapper concept from batch
 			if (exit != null) {
-				if (exit == 0) {
```
ericbottard: Thinking out loud here, without having had a chance to test this:
given that the key in the map is the app name, wouldn't this prevent a redeployment of an app?

Shouldn't we rather report a `== 0 `exited process as `unknown` (_i.e._ as if it never exi(s)ted) and a `!= 0` exited process as `failed`?

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
```diff
@@ -159,10 +159,6 @@ public AppStatus status(String id) {
 			}
 		}
 		AppStatus status = builder.build();
-		// Make sure to explicitly undeploy any exited instances that have status undeployed.
-		if (DeploymentState.undeployed.equals(status.getState())) {
```
ilayaperumalg: I think this specific code was the result of handling `exit == 0` case which set the deployment status to `undeployed`.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
ilayaperumalg: @ericbottard If you agree with the above, then this PR is ready for merge.

ericbottard: My concern was that the deployer should be left in the exact same state after a successful undeployment as it was before the app was ever known (hence state == `unknown`). After careful reading of the code, this is ok, as the key is removed from the map after undeployment.
So, greenlight for me. (even tested with SCDF)

ilayaperumalg:
Rebased and merged as 98ffdb78bafd3e60f81608c0c22f1f9ed2fccc3a

Issue #116, opened 2016-10-03
Standardize JVM option settings


sabbyanandan:
As a developer, I'd like to standardize JVM option settings, to it is consistently applied in all the deployer implementations.

dsyer:
After some back channel chat it seems that for the local deployer we need `JAVA_OPTS` to be `spring.cloud.deployer.javaOpts` (and/or `*.java-opts` via relaxed binding) in `AbstractLocalDeployerSupport`. Then dataflow apps can accept deployment options `app.<name>.spring.cloud.deployer.javaOpts` and the global defaults can be set as `spring.cloud.dataflow.applicationProperties.stream.spring.cloud.deployer.javaOpts` in the dataflow local server app.

markpollack:
Not sure how we can generally javaOpts at the deployer level since for example,  wrt to JVM memory in PCF you can't set the ones you are probably settings now related to memory.

https://github.com/cloudfoundry/java-buildpack/blob/master/docs/framework-java_opts.md#memory-settings.

There is another story for bringing spring.cloud.deployer.memory and disk to the top level

https://github.com/spring-cloud/spring-cloud-dataflow/issues/658

which might be more portable, but have different meanings on different impls, e.g. it could refer to max JVM memory in the local deployer case (via java-opts) but overall container memory for other platforms.

markfisher:
+1 for having `spring.cloud.deployer.memory` map to `-X` args in the local deployer case (as opposed to relying on `javaOpts`, even for local)

markpollack:
For the app-starters the docker file we create is not taking JAVA_OPTS into account, we should improve the dockerfile to have `ENTRYPOINT exec java $JAVA_OPTS -jar /myapp.jar`

spencergibb:
The original `JAVA_OPTS` PR was for java system properties so STS could communicate with spring cloud cli deployed apps.

dsyer:
It has't made it into a release yet though, has it, so can we change it to `javaOpts` at least for 1.0.x?

PR #139, opened 2016-10-28
GH-116 Fix usage of javaOpts

markpollack:
- Change LocalDeployerProperties prefix to spring.cloud.deployer.local
- Add javaOpts property to LocalDeployerProperties, used if no javaOpt
  deployment properties were set in the ApplicationDeploymentRequest
- Support setting the -Xmx Java options using
  spring.cloud.deployer.local.memory in deployment properties of
  ApplicationDeploymentRequest
- Support setting Java Options using
  spring.cloud.deployer.local.javaOpts in deployment properties of
  ApplicationDeploymentRequest
- Switch to use Java 1.8
- Fixed usage of spring.cloud.deployer.local.main and
  spring.cloud.deployer.local.classpath
- Added tests

Resolves spring-cloud/spring-cloud-deployer#116

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
@@ -16,20 +16,18 @@
 
 package org.springframework.cloud.deployer.spi.local;
 
-import java.io.IOException;
 import java.net.URL;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
-import org.springframework.core.io.Resource;
 import org.springframework.http.ResponseEntity;
 import org.springframework.util.Assert;
-import org.springframework.util.StringUtils;
 import org.springframework.web.client.ResourceAccessException;
 import org.springframework.web.client.RestTemplate;
 
@@ -41,6 +39,8 @@ import org.springframework.web.client.RestTemplate;
  * @author Mark Fisher
  */
 public abstract class AbstractLocalDeployerSupport {
+
+	private static final Logger logger = LoggerFactory.getLogger(LocalAppDeployer.class);
 
 	private final LocalDeployerProperties properties;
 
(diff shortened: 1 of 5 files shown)


ilayaperumalg:
When building this PR locally, I see this error from `animal-sniffer` maven plugin:

```
[INFO] --- animal-sniffer-maven-plugin:1.14:check (enforce-java-7) @ spring-cloud-deployer-local ---
[INFO] Checking unresolved references to org.codehaus.mojo.signature:java17:1.0
[ERROR] /spring-cloud/spring-cloud-deployer/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/ExecutionCommandBuilder.java:57: Undefined reference: java.util.stream.Stream java.util.stream.Stream.of(Object[])
[ERROR] /spring-cloud/spring-cloud-deployer/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/ExecutionCommandBuilder.java:57: Undefined reference: boolean java.util.stream.Stream.anyMatch(java.util.function.Predicate)
```

ilayaperumalg:
Looks like the `spring-cloud-build` [forces] (https://github.com/spring-cloud/spring-cloud-build/blob/master/pom.xml#L605) to check for JDK 7: 

```
<plugin>
						<groupId>org.codehaus.mojo</groupId>
						<artifactId>animal-sniffer-maven-plugin</artifactId>
						<version>1.14</version>
						<configuration>
							<signature>
								<groupId>org.codehaus.mojo.signature</groupId>
								<artifactId>java17</artifactId>
								<version>1.0</version>
							</signature>
						</configuration>
						<executions>
							<execution>
								<id>enforce-java-7</id>
								<phase>test</phase>
								<goals>
									<goal>check</goal>
								</goals>
							</execution>
						</executions>
					</plugin>

```

ilayaperumalg:
I think we can just add `.jdk8` file under `spring-cloud-deployer-local` to avoid this check getting activated (since we wanted to compile against jdk 8 here).

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
```diff
@@ -42,6 +40,8 @@
  */
 public abstract class AbstractLocalDeployerSupport {
 
+	private static final Logger logger = LoggerFactory.getLogger(LocalAppDeployer.class);
```
ilayaperumalg: Instead of setting `LocalAppDeployer` class name here, can we have something like this:
```
protected final Logger logger = LoggerFactory.getLogger(getClass());
```

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
```diff
@@ -28,10 +28,15 @@
...
  * @author Ilayaperumal Gopinathan
  */
-@ConfigurationProperties(prefix = "deployer.local")
+@ConfigurationProperties(prefix = LocalDeployerProperties.LOCAL_PROPERTIES)
 public class LocalDeployerProperties {
 
 	/**
+	 * Top level prefix for local deployer configuration properties.
+	 */
+	public static final String LOCAL_PROPERTIES = "spring.cloud.deployer.local";
```
ilayaperumalg: Given this is always referred using `LocalDeployerProperties`, we can probably name this as `PREFIX`.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/ExecutionCommandBuilder.java
```diff
@@ -0,0 +1,100 @@
...
+public class ExecutionCommandBuilder {
+
+    private static final String MAIN =  "main";
+
+    private static final String CLASSPATH = "classpath";
+
+    public void addJavaOptions(List<String> commands, Map<String, String> deploymentProperties,
+                               LocalDeployerProperties localDeployerProperties) {
+        String memory = null;
+        if (containsKey(deploymentProperties, "memory")) {
```
ilayaperumalg: Since only the properties that are prefixed with `spring.cloud.deployer` is considered as `deployment` properties, I think we need to be sure if we want to support `memory` property as both `deployment` property as well as AppDefinition property. An example of this could be from dataflow,

memory as a `deployment` property:
```
dataflow:> stream deploy --name ticktock --properties "app.time.spring.cloud.deployer.local.memory=2048"
```
memory as an app definition property:
```
dataflow:> stream deploy --name ticktock --properties "app.time.memory=2048"
```
It would be convenient to consider supporting both the ways by checking `memory` as a property exists as both `deployment` as well as `definition` property.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
markpollack: cut-n-paste error. sounds good.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
markpollack: Sure!

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/ExecutionCommandBuilder.java
markpollack: This is part of the story that Thomas and I were working on, to add memory and cpu as top level properties.  We should merge this wrt to what we have now and then once Thomas' PR is ready, we can circle back to make the appropriate changes.  That is why I wanted this to get reviewed/merged early as possible wrt to coordinating that work.

ilayaperumalg:
Rebased and merged as e582baf93c671e032eb167072b1cf50c57b1756d along with some updates 5e94b339206ded152e80cc3cabf1f83b0f520ee3

PR #142, opened 2016-11-01
Adds `memory` and `cpu` property constants for SPI AppDeployer.

### LINKED ISSUE #69: Standardize keys for memory, cpu, and disk properties
Memory and Disk size are common properties to set when deploying apps.  They are currently handled by individual deployer implementations defining their own keys.  Making these properties as part of AppDeployer interface will make future additions to Data Flow that enable individual application memory/disk deployment properties be aligned with the code that implements application instance count.  That is, to support 

```
dataflow:> stream deploy --name ticktock --properties "app.time.memory=2048"
```

much like we support 

```
dataflow:> stream deploy --name ticktock --properties "app.time.count=3"
```

See https://github.com/spring-cloud/spring-cloud-dataflow/issues/658

trisberg:
Resolves #69

### DIFF
--- a/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppDeployer.java
+++ b/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppDeployer.java
@@ -28,6 +28,7 @@ import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
  * @author Patrick Peralta
  * @author Marius Bogoevici
  * @author Janne Valkealahti
+ * @author Thomas Risberg
  */
 public interface AppDeployer {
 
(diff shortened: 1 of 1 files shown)


spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppDeployer.java
```diff
@@ -66,6 +67,21 @@
 	static final String INSTANCE_INDEX_PROPERTY_KEY = "INSTANCE_INDEX";
 
 	/**
+	 * The deployment property for the memory setting for the container that will run the app.
+	 * The memory is specified in megabytes with optional trailing 'm' like '1024m'. How this property
```
ilayaperumalg: From this PR: https://github.com/spring-cloud/spring-cloud-deployer/pull/139, `memory` option correlates to JVM max heap size `-Xmx`. But,
>  The memory is specified in megabytes with optional trailing 'm' like '1024m'
this would conflict with the JVM heap size as the default is specified as `bytes` for the heap size.

ilayaperumalg:
LGTM, rebased and merged as 3617e2fa5e7a1674ae91c06c81b03f422d769f3f

PR #143, opened 2016-11-02
Add cleanup(String id) and destroy(String appName) to TaskLauncher interface

### LINKED ISSUE #67: Add cleanup(String id) and destroy(String appName) to TaskLauncher interface
It would be nice to have a `public void cleanup(String id)` and `public void destroy(String appName)` method in the `TaskLauncher` SPI interface since many/most runtimes would need to clean up whatever is created to launch a task.

trisberg:
Resolves #67


markfisher:
LGTM; rebased, resolved conflicts and merged as https://github.com/spring-cloud/spring-cloud-deployer/commit/7bdc1b6d6e70087f7c14379ad84e801a038ae7bd and https://github.com/spring-cloud/spring-cloud-deployer/commit/f36da31cfa7916620e70b6ff0126e67f866ab5f2

Issue #120, opened 2016-10-06
Launch local apps with the same java executable that the deployer used


dsyer:
Users can get a nasty surprise if the java command that is resolved by default from the deployer app is not the same one that it is running with (typically I'm running with Java 8, and my apps are compiled with Java 8, but they fail to launch because the OS defaults to Java 7). It should be easy enough to identify the JDK/JRE using `java.home`.

dsyer:
I just checked: it is sufficient in most cases to simply add "PATH" to the list of `envVarsToInherit` in `LocalDeployerProperties`. If there's some reason not to do that, maybe we could make a special case for the java executable.

markpollack:
I wonder why we didn't add everything returned from System.getenv() as the default.  @markfisher Do you remember the reasoning at the time when Patrick was writing this?

dsyer:
I think a whitelist is better for interprocess comnunication (exporting all of env can lead to leaking credentials for instance). But the java executable in particular could use a different default.

ericbottard:
Actually I wrote this initially, and the rationale for not inheriting the whole environment was that there is no reason the spawned app inherits config (especially boot-style config) from the server. This was even more true at the time of the module launcher.

Still definitely in favor of a whitelist approach (but didn't we change that? the forked process env used to be `clear()`'ed), and also I agree the best default for the java executable is the one that was used to run the server itself.

PR #148, opened 2016-11-02
GH-120 Add PATH to list of inherited env-vars in launched process

markpollack:
Resolves spring-cloud/spring-cloud-deployer#120


markpollack:
Fixes #120

PR #152, opened 2016-11-07
Add default TaskLauncher methods

### LINKED ISSUE #151: Add default TaskLauncher methods
As new `cleanup` and `destroy` methods will break compatibility it would make sense to use default no-op methods on an interface level as some deployer impls will neven need to implement any real functionality.

jvalkeal:
This PR should either get merged or deleted.

- Default interface methods for cleanup
  and destroy. These are no-op methods as
  most impls simply wont need this behaviour.
- Fixes #151


ericbottard:
Are we JDK8+ across the board now?

jvalkeal:
I suppose because deployer is compiled with jdk8 flag.

trisberg:
Kubernetes and Mesos deployers are

markfisher:
LGTM; merged

Issue #89, opened 2016-07-05
List maven repositories that were used when resolution of artifact fails


markpollack:
the error message

```
dataflow:>stream create --name http2file --definition "http | mytransformer | log" --deploy
Command failed org.springframework.cloud.dataflow.rest.client.DataFlowClientException: failed to resolve MavenResource: org.springframework.cloud.stream.app:log-sink-rabbit:jar:1.0.0.BUILD-SNAPSHOT

failed to resolve MavenResource: org.springframework.cloud.stream.app:log-sink-rabbit:jar:1.0.0.BUILD-SNAPSHOT

org.springframework.cloud.dataflow.rest.client.DataFlowClientException: failed to resolve MavenResource: org.springframework.cloud.stream.app:log-sink-rabbit:jar:1.0.0.BUILD-SNAPSHOT
```

should have more info on remote repo locations.

markfisher:
Do you have the full trace? As far as I know the repository URL is listed in the trace, but not exactly where you would expect.

ghillert:
Some discussion points:

The PR is at best a work-around. Ideally the deployer's resource resolution should have proper (abstracted) exception handling defined. 

At present the Deployer SPI does not define exception handling well. We use the super generic `org.springframework.core.io.Resource` class backed by the generic `ResourceLoader` so that everything is abstracted away. Unfortunately, those classes don't define exception handling.

PR #153, opened 2016-11-08
gh-89 List maven repositories that were used when resolution of artifact fails

ghillert:
Resolves https://github.com/spring-cloud/spring-cloud-deployer/issues/89


markpollack:
closed and picked up in https://github.com/spring-cloud/spring-cloud-deployer/pull/154

PR #154, opened 2016-11-08
Add ResourceNotFoundException to DelegatingResourceLoader

### LINKED ISSUE #89: List maven repositories that were used when resolution of artifact fails

markpollack:
Fixes #89

### DIFF
--- a/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
+++ b/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
@@ -137,12 +137,10 @@ public class DelegatingResourceLoader implements ResourceLoader, ResourceLoaderA
 				return new FileSystemResource(cachedResource);
 			}
 		}
-		catch (URISyntaxException e) {
-			throw new IllegalArgumentException(e);
+		catch (Exception e) {
+			throw new ResourceNotFoundException(e.getMessage(), e);
 		}
-		catch (IOException e) {
-			throw new IllegalStateException(e);
-		}
+
 	}
 
 	@Override
@@ -187,4 +185,5 @@ public class DelegatingResourceLoader implements ResourceLoader, ResourceLoaderA
 		}
 		return false;
 	}
+
 }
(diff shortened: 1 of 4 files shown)


markpollack:
Added to https://github.com/spring-cloud/spring-cloud-deployer/pull/153

ghillert:
@markpollack How am I supposed to get access to the Maven repos from the DelegatingResourceLoader? The original intent of the issue was to report Maven repos as part of the exception.

spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java
```diff
@@ -137,12 +137,10 @@ public Resource getResource(String location) {
 				return new FileSystemResource(cachedResource);
 			}
 		}
-		catch (URISyntaxException e) {
```
ghillert: do you think it is acceptable to throw a `
`ResourceNotResolvedException` for a potentially thrown `URISyntaxException`? (I guess it should never happen but ...

markfisher: IMO that should produce an `IllegalArgumentException` that wraps the URI syntax Exception

markpollack: I kept that part of your PR but somehow did not push it in my PR :(.

In my mind the general issue is to have an exception that is specifically for when this method fails, no matter the reason, otherwise we are back to the contract just being totally wide open.  

I'll push an updated PR with the changes you had in it.. I must have screwed up something there.  I'll also do the rename to ResourceNotResolved.

markfisher: If the Exception is going to be applicable more generally than "not-found" problems, then it should probably be renamed to something that is similarly general like `ResourceLoadingException`. Although at that point, I think it's less convincing that adding a new Exception is helpful.

ghillert:
LGTM merging.

- Adding Author tag
- Make sure we use tabs instead of whitespace

ghillert:
Merged with commit 5ca235624acad610c2baa596a00c03a54cac31eb

PR #165, opened 2016-11-25
Add deployer name suffix constant to interface

### LINKED ISSUE #130: Support multiple deployer implementations on the classpath
Some applications would like to have multiple deployer impls on the classpath which is not currently possible.  One approach to support this is to give each deployer bean a unique name so that one can have Spring dependency inject a Map<String,Deployer> of the implemetatations.  There maybe some tweeking of conditional annotations needed.

ilayaperumalg:
- These suffix constants can be used to assign bean names across deployer implementations

Resolves #130


ilayaperumalg:
Once this approach is finalised (based on the `local` deployer PR:https://github.com/spring-cloud/spring-cloud-deployer-local/pull/12), we can add the similar qualifier bean names for other deployer implementations as well.

ilayaperumalg:
@markfisher I agree. Closing this PR and updated the `local` deployer PR with the names updated directly (Inferring from the class name inside the @Bean annotation looks not possible).

Issue #131, opened 2016-10-17
Manage .m2 cache size


sabbyanandan:
As a developer, I'd like to come up with a design to cleanup local maven cache, so it doesn't accumulate to remain unmanaged.

markpollack:
Changing the aether code, while doable, is more difficult than just adding some code into the one entry point we have into that library to perform some manual (java.io.File style) clean up.  This blog post show some of the code that we will likely need to add in order to apply some strategies for cleanup (delete oldest first etc)

http://blog.trifork.com/2011/08/01/cleaning-up-your-maven-repository/

ericbottard:
The real problem IMHO is that we don't control the scope of the resolved MavenResource.
When is it safe to delete files?

markpollack:
This issue probably belongs in the CF deployer project.

There is only one entry point into the aether library to download the artifact to the local disk.  That is happening inside `CloudFoundryAppDeployer.pushApplication` - the call to `getApplication` will call `request.getResource().getFile().toPath();` and that is what does the resolution/download.  

https://github.com/spring-cloud/spring-cloud-deployer-cloudfoundry/blob/master/src/main/java/org/springframework/cloud/deployer/spi/cloudfoundry/CloudFoundryAppDeployer.java#L284

If we surround this method with an appropriate lock based on a fully qualified name of the resource, then right after the app is pushed to CF, we can do cleanup of that specific artifact.  For example, we can look to remove all older versions of the app that was just deployed and maybe fallback to deleting the app itself is disk space is still low.

gregturn:
It may be tricky to keep inside the scope of the SPI, since `deploy` and `status` are separate operations.

I'd argue that the user should invoke the cleanup after `status` shows the proper results.

As an example, this is how I'm using the CF Deployer. I definitely plan to add a cleanup step after the highlighted line (wrapped in a poller).

https://github.com/gregturn/clouddriver/blob/reactor/clouddriver-cf/src/main/groovy/com/netflix/spinnaker/clouddriver/cf/deploy/handlers/CloudFoundryDeployHandler2.groovy#L158

markpollack:
I don't understand how `status` enters into this discussion of disk mgmt.  I'm hoping that keeping just the latest version of each .jar file around will be efficient enough on the disk usage.  If a deployed app succeeds or fails that does not seem to be a concern to how we manage the disk usage.

markpollack:
[...]

PR #166, opened 2016-12-12
Manage maven cache size

ericbottard:
Introduce a wrapper around an existing ResourceLoader that limits the
total size used on disk, purging LRU entries first.

Fixes https://github.com/spring-cloud/spring-cloud-deployer/issues/131


spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/EOFInputStream.java
```diff
@@ -0,0 +1,82 @@
+package org.springframework.cloud.deployer.resource.support;
```
ilayaperumalg: Missing copyright.

spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/EOFInputStream.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.io.FilterInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Created by ericbottard on 08/12/16.
```
ilayaperumalg: Would be a good idea to change this in your IDE to have `@author` etc.,

spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/LRUResourceLoader.java
```diff
@@ -0,0 +1,99 @@
+package org.springframework.cloud.deployer.resource.support;
```
ilayaperumalg: copyright.

ericbottard:
Added copyright where appropriate

jvalkeal:
I don't think this will work in a real life:

1. `Resource` may not have a backing physical `File`
2. With maven repo you should purge all files related to an artifact(pom,sha, etc).
3. Having multiple caches in a same system will start to fight with each others, right?

ericbottard:
1. This is meant to wrap our DelegatingResourceLoader, so each resource *will* be a File
2. True in theory, but a) this may apply to non-maven resources. b) we only want to avoid crash in resource constraint envs, not be 100% perfect maven-wise. There may be a problem with timestamped SNAPSHOTs though.
3. I don't get what you're saying

ericbottard:
I know this is not an ideal solution, serves more as a basis for discussion (which already started with @markpollack )

jvalkeal:
Yeah, this is a bit complex issue. I don't want any external system to remove stuff from my local m2 repo, that's way too invasive. Back in days there used to be a lot of issues with gradle if jar artifact was removed but metadata files told file is there. What about if that file in currently in use by some other process?

Maybe this should not be a public impl?

ericbottard:
This is mainly for use in CF, not for a "laptop" m2 cache

jvalkeal:
ah, ok. I surely don't care about m2 on CF(until someone starts to use that dave's thin wrapper).

spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/LRUResourceLoader.java
```diff
@@ -0,0 +1,115 @@
...
+			super(5, 0.75f, true);
+			this.maxSize = maxSize;
+		}
+
+
+		@Override
+		protected boolean removeEldestEntry(Map.Entry<File, Long> eldest) {
+			// Compute cumulative used disk space, in MRU order
+			List<Map.Entry<File, Long>> lruLast = new ArrayList<>(this.entrySet());
+			Collections.reverse(lruLast);
```
markpollack: why reverse the entries as they are stored by insertion order?

ericbottard: They're actually stored in last access order (LRU first), see 3rd parameter to the constructor (and associated javadoc). So I want them in reverse order to have the bigger sums represent the LRU items

markpollack:
1. We can inforce the impl to only deal with files to make it explicit.  We can also just put this as a protected class bandaid in the cf-deployer code base to limit the scope of usage since it is only for CF. ATM
2. It would seem deleting everything under the .m2 version directory to remove all maven artifacts would be the cleaner approach.
3. " I don't want any external system to remove stuff from my local m2 repo", well you do care if it means the SCDF Server on PCF can only run for a couple weeks before running out of disk space.

spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/LRUResourceLoader.java
```diff
@@ -0,0 +1,115 @@
...
+	private final ClassLoader classLoader = ClassUtils.getDefaultClassLoader();
+
+	private final ResourceLoader delegate;
+
+	private final Map<File, Long> cache;
+
+	/**
+	 * Wrap an existing ResourceLoader, limiting total used space to {@literal maxSize} bytes.
+	 */
+	public LRUResourceLoader(ResourceLoader delegate, long maxSize) {
```
markpollack: I don't see how we can know the size of this specific file up front to set it to an appropriate value.  Consider the case where we have 1GB of free disk.  I should be able to have a 500MB .jar file and 10 * 50MB other artfiacts or have 100 * 50 MB file.  How to set the maxSize in this case?

ericbottard: maxSize here is the total combined cache size. So set it to 1GB in your case

markpollack:
Here is an alternative implementation after a google hangout with Eric that should help solve the issues of snapshot and also proper cleanup of directories (vs. just files)

1. We use the config property ${maven.localRepository} to determine the path where the maven cache is located.  This will later be used to determine the free disk space on the underlying partition for that path.
2. The LRU cache will only keep track of filenames that were most recently used, no bookkeeping of file size.  This is due to poor bookkeeping when mvn downloads snapshots - there is an extra copy of the latest file even when just one snapshot is downloaded and the bookkeeping increases inaccuracy when there are older snapshots laying around.
3. As a ctor argument we specify the low water mark of free disk space that will trigger the cleanup operation.  This will be expressed as a percentage, default can be 10%.  
4. After a resource is downloaded, we check if the low water mark of free disk space has been exceeded, iterate through the LRU cache and delete the parent directory of the LRU files until the low water mark is not exceeded.
5. Move this this to spring-cloud-dataflow-server-cloudfoundry repo and provide a new instance of DelegatingResourceLoader to override what is provided by `DataFlowControllerAutoConfiguration` (there is a `@ConditionalOnBean` annotation in the autoconfig of the dataflow core.

ericbottard:
Closing, will open a PR against https://github.com/spring-cloud/spring-cloud-dataflow-server-cloudfoundry

PR #171, opened 2017-01-04
Make TaskStatus.getAttributes public

### LINKED ISSUE #168: TaskStatus.getAttributes() is not public and there's no indication why not
Maybe if there's a reason it's not public, it could be mentioned in the javadoc? It's quite hard to work with status objects as the LaunchState changes unless the attributes are public.

trisberg:
Resolves #168


markpollack:
merged as 599710babfc1805a284a4d1803532a2d3fd1a6c4

PR #179, opened 2017-01-17
Add URI validation for URIRegistryPopulator

### LINKED ISSUE #176: UriRegistryPopulator should not ignore exception for invalid app names
The UriRegistryPopulator should not ignore IllegalArgumentException [1] when populating the registry from a properties file. This can lead to invalid data getting registered causing problems for users of the app registry.

[1] https://github.com/spring-cloud/spring-cloud-deployer/blob/2c2a0d18cf7271d349a2517aed53ea4f7834edb2/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/registry/UriRegistryPopulator.java#L85

ilayaperumalg:
- logs error messages where there are URI validation errors
 - Add test

Resolves #176

### DIFF
--- a/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/registry/UriRegistryPopulator.java
+++ b/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/registry/UriRegistryPopulator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
Will update the PR based on the discussion in #176 on handling `IllegalArgumentExçeption` and updating JavaDoc.

trisberg:
Do we need this validation? We do this same validation in the implementation for the RdbmsUriRegistry now. I think it is better to leave extra validation to implementations.

ilayaperumalg:
I was thinking we can have the common validation at the top level as the implementations may/may not have the validation.

trisberg:
Merged as commit 1abf9a43bd15c2d5a8f8db10d91526c07033b697

Fixed the warning message formatting and added test for empty URI

Issue #181, opened 2017-01-27
UriRegistryPopulator should accept a Resource/InputStream


ericbottard:
Code is currently duplicated between UriRegistryPopulator and dataflow.
UriRegistryPopulator should accept a Resource (and not a resource location), then dataflow can just pass the inputStream to the properties "file".

PR #182, opened 2017-01-30
UriRegistryPopulator now takes a Resource. Use static access

ericbottard:
Fixes https://github.com/spring-cloud/spring-cloud-deployer/issues/181

Needed by https://github.com/spring-cloud/spring-cloud-dataflow/pull/1145


ericbottard:
Issued a PR for the driving issue https://github.com/spring-cloud/spring-cloud-dataflow/pull/1145

ericbottard:
After discussion with Mark, we came to the conclusion that this "breaking" change is very unlikely to break any consumer really. Moreover, that class may very well move to another project / package soon, and should not be considered part of the API of the deployer project.
Marked the class as deprecated, we decided to move forward with the PR

ilayaperumalg:
LGTM, merging.

PR #198, opened 2017-03-06
Add environmentInfo method to API

### LINKED ISSUE #183: Add environment/platform info API
As a user, I'd like to call an API to gather platform info, so I could use it to probe and aggregate all runtime details. 

This API, depending on the platform where it's invoked from, it could potentially give us these details: https://github.com/spring-cloud/spring-cloud-dataflow/wiki/Reporting-Issues

trisberg:
- add utility class to generate a base version for environment info

- add test to verify info contains required entries

Resolves #183

### DIFF
--- a/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppDeployer.java
+++ b/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppDeployer.java
@@ -16,6 +16,8 @@
 
 package org.springframework.cloud.deployer.spi.app;
 
+import java.util.Map;
+
 import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
 
 /**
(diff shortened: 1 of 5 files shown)


spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppDeployer.java
```diff
@@ -135,4 +137,13 @@
...
 	AppStatus status(String id);
+
+	/**
+	 * Return the environment info for this deployer. This info should include the
+	 * following: 'deployer-class-name','deployer-version', 'deployer-spi-version', 'platform-type',
+	 * 'platform-client-version', 'platform-host-version', 'java-version', 'spring-version', 'spring-boot-version'.
+	 *
+	 * @return the app deployment environment info
+	 */
+	Map<String, String>  environmentInfo();
```
ericbottard: I'd +1.

Note that `deployer-classname` is something that 
1) is not very interesting to see for an end-user once you have the other properties
2) can be retrieved by doing `getClass()` on the very object implementing this interface

trisberg: OK, let me add a commit with a new return type, then we can argue what should be included.

I'll start with:

- deployerSpiVersion
- deployerImplementationVersion
- platformType
- platformClientVersion
- platformHostVersion
- javaVersion
- springVersion
- springBootVersion

ghillert:
@trisberg I think we also need a Map<String,String> property in `DeployerEnvironmentInfo` for _optional_ or _platform specific_ properties. I would think that e.g. the CF implementation would want to add a slur of additional values.

ghillert:
LGTM - merging.

ghillert:
Merged with commit 076512c8aa90bfbc32e18034494564c06e5b595f

PR #216, opened 2017-05-11
GH-215 added toString() implementation

### LINKED ISSUE #215: Add toString() implementation to AppDeploymentRequest
Would be rather helpful while debugging

olegz:
- added toString() Implementation to AppDeploymentRequest
- polished/simplified some code

Resolves #215

### DIFF
--- a/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/core/AppDeploymentRequest.java
+++ b/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/core/AppDeploymentRequest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.Map;
 
 import org.springframework.core.io.Resource;
+import org.springframework.core.style.ToStringCreator;
 import org.springframework.util.Assert;
 
 /**
@@ -43,6 +44,7 @@ import org.springframework.util.Assert;
  *
  * @author Mark Fisher
  * @author Janne Valkealahti
+ * @author Oleg Zhurakousky
  */
 public class AppDeploymentRequest {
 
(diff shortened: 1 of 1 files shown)


spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/core/AppDeploymentRequest.java
```diff
@@ -92,12 +94,8 @@ public AppDeploymentRequest(AppDefinition definition, Resource resource,
 		Assert.notNull(resource, "resource must not be null");
 		this.definition = definition;
 		this.resource = resource;
-		this.deploymentProperties = deploymentProperties == null
-				? Collections.<String, String>emptyMap()
-				: Collections.unmodifiableMap(deploymentProperties);
-		this.commandlineArguments = commandlineArguments == null
-				? Collections.<String>emptyList()
-				: Collections.unmodifiableList(commandlineArguments);
+		this.deploymentProperties = deploymentProperties;
```
ilayaperumalg: Isn't it still be possible to have `null` values for `deploymentProperties` and `commandlineArguments` when this constructor is invoked `directly` with `null` values?

olegz: Well if you look at the existing code the `null` was replaced with empty `List/Map`, so I've just eliminated the unnecessary IF

ilayaperumalg: But, what happens if this constructor `public AppDeploymentRequest(AppDefinition definition, Resource resource,Map<String, String> deploymentProperties, List<String> commandlineArguments)` is 
`directly` invoked with `null` deploymentProperties and commandlineArguments ?

olegz: @ilayaperumalg  I see your point, let me revise it a bit

olegz:
@ilayaperumalg I've reverted constructor changes keeping `toString()` only. Let me know if you want me to squash it

ilayaperumalg:
@olegz I can squash them during the merge.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed and merged as 47b2f6f

PR #221, opened 2017-06-07
GH-220 removed Resource requirement

### LINKED ISSUE #220: AppDeploymentRequest should not require Resource
AppDeploymentRequest should not require Resource in the event when deployment properties provide both `main` and `classpath`. 
For example, in the case of Local Deployer the below code snippet would allow a successful deployment without resource
```
deploymentProperties.put(LocalDeployerProperties.PREFIX + ".main", "sample.SampleProgram");
deploymentProperties.put(LocalDeployerProperties.PREFIX + ".classpath", new ClassPathResource("").getFile().getAbsolutePath());
```

Basically it creates an inconsistency with the deployer and what it expects

olegz:
- removed 'required' assertion for Resource
- added explanation to javadoc as to why Resource can be null

Resolves #220

### DIFF
--- a/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/core/AppDeploymentRequest.java
+++ b/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/core/AppDeploymentRequest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


ericbottard:
Not sure I understand the actual usecase where this could really be null. Are you referring to a pure bash/shell job?
If so, isn't that such a special case than we could require something like a `NullResource`?

olegz:
@ericbottard 
I am not sure that is a special case. 
Let me try it this way.  . .
I have a microservice which is not a boot app, so in order for it to be _deployed_ using my "custom and amazing deployer implementation", I may provide additional instructions via deployment properties which will help me to construct a Java command within the mentioned deployer implementation (e.g., `java -cp /foo/bar. . . foo.bar.MyMain`). Local Deployer is actually one that supports that already and does work with dummy Resource - providing `main` and `classpath` properties are provided via )deploymentProperties_.
Sure, we can extract NullResource from tests and make it public, and use it as suggested. However, IMHO I do not think it is a good and/or clean programming model. In fact I would go as far as removing `Resource` as a constructor argument all together and supply it as a deployment property simply because `Resource` *only* makes sense to the _deployers_ that are _Resource-aware_ so to speak, but that may be a breaking change.
WDYT?

ericbottard:
But where would the classes (and supporting classes, ie "the classpath") of your app come from?
They need to come from someplace, and each __deployed__ app is different.

olegz:
I am probably missing your point, so let me paste a piece of code that demonstrates what I mean with current functionality of the Local Deployer. Hopefully it will bring us on the same page ;) Basically you supply the classpath
```java
LocalAppDeployer deployer = new LocalAppDeployer(new LocalDeployerProperties());
Map<String, String> properties = new HashMap<>();
AppDefinition definition = new AppDefinition("foo", properties);

ClassPathResource resource = new ClassPathResource(""); // dummy resource

Map<String, String> deploymentProperties = new HashMap<>();
deploymentProperties.put(LocalDeployerProperties.PREFIX + ".main", "foo.bar.MySampleMain");
deploymentProperties.put(LocalDeployerProperties.PREFIX + ".classpath", new ClassPathResource("").getFile().getAbsolutePath());

AppDeploymentRequest request = new AppDeploymentRequest(definition, resource, deploymentProperties);
deployer.deploy(request);
```
In the above sample, you can see that I am pointing to my current classpath, but could easily point to anything else (JARs, directory etc.). And I explicitly provide the main class.

olegz:
After discussing it with @ericbottard we agreed that while this is definitely a bug, the proposed approach to fix does not address the bigger issue which goes directly into the scope of the "deployer" and that it is NOT a generic java command executor, rather a deployer for artifacts. So even in cases of non-boot apps the deployable artifact could be ZIP, directory or anything else, but it must contain (in its definition) all that is necessary to deploy the application. So essentially such artifact must be described thru Resource abstraction.

PR #233, opened 2017-10-03
Add deployer resource command

shahbour:
I added these to support sending simple command to mesos deployer where i just wanted to run fat jar instead of having to build it on docker .

I used command as schema prefix


markpollack:
I am not sure I understand the use-case.  

>  mesos deployer where i just wanted to run fat jar instead of having to build it on docker .

How does one reference the location of the fat-jar via the command, is there some shared filesystem at play

sabbyanandan:
Closing this due to inactivity.

PR #234, opened 2017-10-09
Add scale to SPI and tests

### LINKED ISSUE #62: Add APIs to scale apps up/down
As a developer, I'd like to add the `scale()` API, so the platform deployers can implement the platform-specific approach to scale app instances up/down.

dturanski:
Resolves #62

### DIFF
--- a/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppDeployer.java
+++ b/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppDeployer.java
@@ -30,6 +30,7 @@ import org.springframework.cloud.deployer.spi.core.RuntimeEnvironmentInfo;
  * @author Marius Bogoevici
  * @author Janne Valkealahti
  * @author Thomas Risberg
+ * @author David Turanski
  */
 public interface AppDeployer {
 
@@ -143,4 +144,12 @@ public interface AppDeployer {
 	 * @return the runtime environment info
 	 */
 	RuntimeEnvironmentInfo environmentInfo();
+
+	/**
+	 * Scale an app to the number of replicas.
+	 *
+	 * @param id the app deployment id, as returned by {@link #deploy}
+	 * @param replicas the desired number of replicas
+	 */
+	void scale(String id, int replicas);
 }
--- a/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppDeployer.java
+++ b/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppDeployer.java
@@ -146,10 +146,10 @@ public interface AppDeployer {
 	RuntimeEnvironmentInfo environmentInfo();
 
 	/**
-	 * Scale an app to the number of replicas.
+	 * Scale an app to the desired count.
 	 *
 	 * @param id the app deployment id, as returned by {@link #deploy}
-	 * @param replicas the desired number of replicas
+	 * @param desiredCount the desired number of instances.
 	 */
-	void scale(String id, int replicas);
+	void scale(String id, int desiredCount);
 }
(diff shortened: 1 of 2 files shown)


spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppDeployer.java
```diff
@@ -143,4 +144,12 @@
...
 	 */
 	RuntimeEnvironmentInfo environmentInfo();
+
+	/**
+	 * Scale an app to the number of replicas.
+	 *
+	 * @param id the app deployment id, as returned by {@link #deploy}
+	 * @param replicas the desired number of replicas
+	 */
+	void scale(String id, int replicas);
```
sabbyanandan: @markpollack suggests `void scale(String deploymentId, int desiredCount)`

markfisher: I wonder if we should support (instead or in-addition) a +/- N option... otherwise it's the user's responsibility to know the current state before calling this, and that seems to create a race condition

dturanski: @markfisher I'm not sure I follow. If the user would scale, don't they already know the current count? This follows the protocol of k8s and PCF just tell me how many instances you want.

sabbyanandan:
We will revisit in the context of Spinnaker integration in the future.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppDeployer.java
dturanski: @markfisher - I agree with your suggestion now that we are revisiting this,  especially for autoscaling use cases.

markfisher: I'm not so sure about my own suggestion after reading this: https://hackernoon.com/level-triggering-and-reconciliation-in-kubernetes-1f17fe30333d

dturanski: Interesting perspective. I was thinking if we have a monitoring app that can scale instances, it probably behaves more edge triggered. If the app is operating beyond its capacity, then +1, if it is below its capacity then -1.  It may only care how many instances are running at the extremes (e.g. the dial doesn't go to 11, or below 0).  I'm leaning toward providing both . 
```
void scaleAbsolute(String id, int desiredCount);
int scaleRelative(String id, int delta);
``` 
@tzolov ?

jvalkeal: Requesting a new count is easiest way to get started as then you don't get race conditions. We had scaling in our good ol' yarn deployer and at least in a hadoop world you had to set a new count.

markfisher: Kubernetes is the same (as described in the edge vs level triggering article I posted above), and that avoids the race conditions which are exacerbated by the fact that scaling can take some time, during which multiple requests for changes may occur.

tzolov: @dturanski , here is a snipped from my s1p demo. A web hook is configured to receive 2 types of (json) alerts, coming from a Prometheus AlertManager: (1) ThroughtputDeltaHigh - the throughput between the source and processor (e.g. `fraud-detection`)  apps is higher than a given threshold; (2) ThroughtputDeltaZero - the throughputs of the source and the processor apps are the same.  
When the delta is high the `fraud-detection` processor is scaled to 4 instances and reduced to 1 otherwise. I believe this is aligned with the level approach. We express the desired state/level and let the platform reach it eventually. Most of the PID control logic is done though the AlertManager configuration.
 The scale API looks like `scale(streamName, AppName, Count)`

```java
@RequestMapping(value = "/alert", method = RequestMethod.POST)
public void alertMessage(@RequestBody List<Map<String, Object>> alerts) {
   for (Map<String, Object> alert : alerts) {
      String alertStatus =  fn(from alert);
      String alertName =  fn(alert);
      String streamName = fn(alert) ;

      if (alertStatus.equals("firing")) {
         if (alertName.equals("ThroughtputDeltaHigh")) {
            kubeScaleService.scale(streamName, "fraud-detection", 4);
         }  else if (alertName.equals("ThroughtputDeltaZero")) {
            kubeScaleService.scale(streamName, "fraud-detection", 1);
         }
      }
   }
}
```

PR #239, opened 2017-11-08
Add explicit Http resource loader

### LINKED ISSUE #237: Remove cache resources in delegating resource loader
After some discussion related to #235, it is decided to remove the caching of downloaded resources and leave it to the resource loader implementations to handle the caching.

For instance, MavenResource's artifact resolver would use the retrieval mechanism based on repository policies. and, some resource loaders wouldn't want to cache at all.

### LINKED ISSUE #238: Use explicit Http resource loader
The default resource loader's UrlResource doesn't work right for downloading the Http resources for  app deployer as it always resolves to `file:` to download file system resource.

Hence, we need an explicit HttpResourceLoader to download `http` resources when using DelegatingResourceLoader.

ilayaperumalg:
- This will make sure to use HttpResource
 - Add HttpResourceLoader that downloads the URLResource instead of expecting it to be on the File system
 - Add this to delegating resource loader when using schemes `http` and `https`

Resolves #238
Resolves #237

### DIFF
--- a/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/HttpResource.java
+++ b/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/HttpResource.java
@@ -37,22 +37,24 @@ import org.springframework.util.FileCopyUtils;
  */
 public class HttpResource extends UrlResource {
 
-	private final String location;
-
 	public HttpResource(String location) throws MalformedURLException {
 		super(location);
-		this.location = location;
 	}
 
 	@Override
 	public String getDescription() {
-		return "Http Resource [" + location + "]";
+		try {
+			return "Http Resource [" + this.getURL().getPath() + "]";
+		}
+		catch (IOException e) {
+			throw new RuntimeException(e);
+		}
 	}
 
 	@Override
 	public File getFile() throws IOException {
-		String fileName = ShaUtils.sha1(location);
-		File tempFile = new File(Files.createTempDirectory("").toFile(), fileName);
+		String fileName = ShaUtils.sha1(this.getURL().getPath());
+		File tempFile = new File(Files.createTempDirectory("spring-cloud-deployer").toFile(), fileName);
 		FileCopyUtils.copy(this.getInputStream(), new FileOutputStream(tempFile));
 		tempFile.deleteOnExit();
 		return tempFile;
--- a/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/HttpResource.java
+++ b/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/HttpResource.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/HttpResource.java
```diff
@@ -0,0 +1,61 @@
...
+		this.location = location;
+	}
+
+	@Override
+	public String getDescription() {
+		return "Http Resource [" + location + "]";
+	}
+
+	@Override
+	public File getFile() throws IOException {
```
markpollack: There should be some tests for this method.
I don't see why we need SHA anymore, it makes things more opaque.  Similarly, we could group all these files under a directory "spring-cloud-deployer" instead of having them as top level in the temp dir.  This method would also benefit from being synchronized.

markpollack: Also, was reading a bit more into temp files...

https://www.pongasoft.com/blog/yan/java/2011/05/17/file-dot-deleteOnExit-is-evil/

is a quick read.  We should prob remove the deleteOnExit.    We can do the disk mgmt via https://github.com/spring-cloud/spring-cloud-dataflow-server-cloudfoundry/blob/master/spring-cloud-dataflow-server-cloudfoundry-autoconfig/src/main/java/org/springframework/cloud/dataflow/server/cloudfoundry/resource/LRUCleaningResourceLoader.java 

perhaps  `LRUCleaningResourceLoader` needs to move into spring-cloud-deployer...

markpollack: Also, just generally looking over `Resource` implementation in core spring.  We probably need to override some other methods as well, e.g. `getOutputStream`, `isWritable`, `createRelative` so that is is a "true" Resource implementation vs. one that just happens to work for the deployer use case ATM.    I think that we may need to keep the file that is downloaded 'on demand' in a private field, so that other methods can reference it.  For example, getOutputStream should not have to download the file again.

spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/HttpResource.java
```diff
@@ -0,0 +1,61 @@
...
+import org.springframework.util.FileCopyUtils;
+
+/**
+ * A {@link Resource} implementation for resolving HTTP resource.
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+public class HttpResource extends UrlResource {
+
+	private final String location;
```
markpollack: don't need to store this, can call `getURL()` instead.

spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/HttpResource.java
ilayaperumalg: It looks like `getInputStream` always invokes the download of Http resource which I believe is sufficient along with having a new temp file every time getFile is invoked. Do you see any issue with that?

ilayaperumalg: Pushed a test that shows the temp files are distinct. Also, it looks like `deleteOnExit` is unavoidable at least for now as that seems to be the only possible way to delete those temp files. Given, we don't explicitly invoke delete on these files, possibility leakage isn't there I believe.

ilayaperumalg:
This was merged as part of ade190d7b37aaea6978f6bf3485e5cdf4df82828

Issue #235, opened 2017-10-23
Revisit the resource caching requirement


sabbyanandan:
As a user, while deploying a stream using SCDF to K8S, I see the logs regarding caching the docker images, which seems irrelevant for non-maven artifacts. Though we [log the inapplicability](https://github.com/spring-cloud/spring-cloud-deployer/blob/master/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java#L128), it seems like we are still [creating a cache directory](https://github.com/spring-cloud/spring-cloud-deployer/blob/master/spring-cloud-deployer-resource-support/src/main/java/org/springframework/cloud/deployer/resource/support/DelegatingResourceLoader.java#L121), and that is not required. Perhaps we can remove the directory after logging the `WARN` message?

Logs:
```
2017-10-23 15:29:42.866  INFO 80812 --- [nio-7577-exec-3] o.s.c.d.r.s.DelegatingResourceLoader     : Caching resource 'docker:springcloudstream/log-sink-rabbit:1.2.0.RELEASE' as file /var/folders/c3/ctx7_rns6x30tq7rb76wzqwr0000gs/T/deployer-resource-cache3504604345247190928/docker-293360093b9ac0e97da54d2c95043fa08a8d6e2d
2017-10-23 15:29:42.866  WARN 80812 --- [nio-7577-exec-3] o.s.c.d.r.s.DelegatingResourceLoader     : Unable to cache file since getInputStream() is not supported for resource: Docker Resource [docker:springcloudstream/log-sink-rabbit:1.2.0.RELEASE]
2017-10-23 15:29:43.037  INFO 80812 --- [nio-7577-exec-3] o.s.c.d.s.k.DefaultContainerFactory      : Using Docker image: springcloudstream/log-sink-rabbit:1.2.0.RELEASE
2017-10-23 15:29:43.037  INFO 80812 --- [nio-7577-exec-3] o.s.c.d.s.k.DefaultContainerFactory      : Using Docker entry point style: exec
[...]

PR #241, opened 2017-11-12
Create DownloadingUrlResource and LRUCleaningResourceLoader

markpollack:
This contains the commit in https://github.com/spring-cloud/spring-cloud-deployer/pull/239 and builds upon it. This PR resolves spring-cloud/spring-cloud-deployer#235.

The impl in the previous PR did not take into account usage when we add `LRUCleaningResourceLoader`.

When using `LRUCleaningResourceLoader` `resource.getFile` can often be invoked twice, and we need to return the same file for both invocations, otherwise we will not track all downloaded files in the LRU.

e.g in local deployer

```
Resource resource = request.getResource();
try {
	commands.add(resource.getFile().getAbsolutePath());
}
```
and the LRU cache

```
	public Resource getResource(String location) {
		Resource resource = delegate.getResource(location);
		try {
			File file = resource.getFile();
			synchronized (lruCache) {
				lruCache.put(file, null);
			}
			return resource;
		} catch (IOException e) {
			logger.debug("{} is not stored on the local filesystem, skipping", resource);
			return resource;
		}
	}
```

Added some additional tests


ilayaperumalg:
LGTM, merging. Also, verified the changes by running some sample deployments via SCDF/Skipper.

PR #254, opened 2018-07-05
Add global update/checksum policy

### LINKED ISSUE #252: Add support configuring update policy for `local` repository
We currently have support for configuring `update-policy` for remote repositories.
We need to extend this for `local` repository configuration as well.

ilayaperumalg:
- Add update/checksum policies as global level maven properties
 - These properties would be set at the RepositorySystemSession configuration
 - The remote repositories would still have their own RepositoryPolicy override options

Resolves #252


markpollack:
rebased, merged as 3ed4c680c2d27d9f4b02ad84ad2bdf8a4c16dcef

PR #286, opened 2019-03-22
Add methods to TaskLauncher

dturanski:
This PR is related to https://github.com/spring-cloud/spring-cloud-dataflow/issues/2855 
This adds new methods to the `TaskLauncher` interface.  Corresponding PRs to implement these are in the local, cloudfoundry, and kubernetes implementations.

### DIFF
--- a/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/task/TaskLauncher.java
+++ b/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/task/TaskLauncher.java
@@ -110,11 +110,15 @@ public interface TaskLauncher {
 	 *
 	 * @return the maximum concurrent task executions allowed.
 	 */
-	int getMaximumConcurrentTasks();
+	default int getMaximumConcurrentTasks() {
+		throw new UnsupportedOperationException("'getMaximumConcurrentTasks' is not implemented.");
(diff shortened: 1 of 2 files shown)


spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/task/TaskLauncher.java
```diff
@@ -103,4 +104,17 @@
...
+	 *
+	 * @return the maximum concurrent task executions allowed.
+	 */
+	int getMaximumConcurrentTasks();
+
+	/**
+	 *
+	 * @return the count of currently running task executions if known.
+	 */
+	int getRunningTaskExecutionCount();
```
markpollack: Funny, I thought maybe the method name `getCurrentlyRunningTaskExecutionCount` or `getCurrentTaskExecution` would be a slightly better name as it aligns with the javadoc description.  I think all the methods have an implied 'if known' to them.

markpollack: add default method to throw `UnsupportedOperationException`

markpollack:
squashed and merged as e347a6c4957fd0bca217390d15d4bec29bf5e2c5

PR #303, opened 2019-04-05
Add aspect to log MavenArtifactResolver.resolve()

dturanski:
Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/2909

Uses AspectJ to log messages for invocations of `MavenArtifactResolver.resolve()`. Since this is not a Spring bean, Spring AOP doesn't apply. Artifact paths are cached to avoid lots of duplicate log messages.


dturanski:
Closed without merging changes not needed.

PR #304, opened 2019-05-28
Switch to maven resolver from aether

### LINKED ISSUE #245: Upgrade to use https://maven.apache.org/resolver/ instead of old aether-impl
Maven (the build tool) does not directly import the aether-impl jar.  It has created a new project (with the same aether classes) here - https://github.com/apache/maven-resolver

This should be considered the 'defacto standard' for maven artifact resolution.

markpollack:
Fixes #245

### DIFF
--- a/spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/ConsoleRepositoryListener.java
+++ b/spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/ConsoleRepositoryListener.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


markpollack:
Should use Logger instead of direct print

spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/ConsoleRepositoryListener.java
```diff
@@ -0,0 +1,128 @@
...
+	public void metadataResolving( RepositoryEvent event )
+	{
+		println( "metadataResolving", event.getMetadata() + " from " + event.getRepository() );
+	}
+
+	private void println( String event, String message )
+	{
+		out.println( "Aether Repository - " + event + ": " + message );
+	}
+}
```
dturanski: Add new line

dturanski: Agree with using Logger

cppwfs:
With the replacement of aether with maven, the link to https://github.com/eclipse/aether-core/blob/4cf5f7a406b516a45d8bf15e7dfe3fb3849cb87b/aether-api/src/main/java/org/eclipse/aether/repository/RepositoryPolicy.java#L16 in the configuration.adoc from the  SCDF reference documentation needs to be updated.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed and merged as de4930e

PR #316, opened 2019-08-23
Add default method for getLog

### LINKED ISSUE #315: Define default methods for Deployer SPI interfaces
The AppDeployer/TaskLauncher `getLog` methods need to have default methods defined so that the deployer implementations could skip the implementations or use it between the 2.0.x version upgrades.

ilayaperumalg:
- In both the AppDeployer/TaskLauncher SPI interfaces, adda default method for `getLog` which throws the UnsupportedOperationException

 - Add tests

Resolves #315


cppwfs:
LGTM

PR #319, opened 2019-10-16
Add scale to SPI and tests

dturanski:
Fixes #234

### DIFF
--- a/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppDeployer.java
+++ b/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppDeployer.java
@@ -31,6 +31,7 @@ import org.springframework.cloud.deployer.spi.core.RuntimeEnvironmentInfo;
  * @author Janne Valkealahti
  * @author Thomas Risberg
  * @author Ilayaperumal Gopinathan
+ * @author David Turanski
  */
 public interface AppDeployer {
 
@@ -153,4 +154,11 @@ public interface AppDeployer {
 	default String getLog(String id) {
 		throw new UnsupportedOperationException("'getLog' is not implemented.");
 	}
+	/**
+	 * Scale an app to the desired count.
+	 *
+	 * @param id the app deployment id, as returned by {@link #deploy}
+	 * @param desiredCount the desired number of instances.
+	 */
+	void scale(String id, int desiredCount);
 }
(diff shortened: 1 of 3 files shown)


spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppDeployer.java
```diff
@@ -153,4 +154,11 @@
 	default String getLog(String id) {
 		throw new UnsupportedOperationException("'getLog' is not implemented.");
 	}
+	/**
+	 * Scale an app to the desired count.
+	 *
+	 * @param id the app deployment id, as returned by {@link #deploy}
+	 * @param desiredCount the desired number of instances.
+	 */
+	void scale(String id, int desiredCount);
```
jvalkeal: Should we make this a default method same as `getLog`? Then it would not break anything immediately.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppDeployer.java
```diff
@@ -153,4 +154,11 @@
 	default String getLog(String id) {
 		throw new UnsupportedOperationException("'getLog' is not implemented.");
 	}
+	/**
+	 * Scale an app to the desired count.
+	 *
+	 * @param id the app deployment id, as returned by {@link #deploy}
```
tzolov: Shouldn't this be  the `AppDefinition Name` not the `deployment Id` (assuming the later represents the the app instance id).  I believe in the tests above the correct AppDef name is used. 
As i see it, the API should signal a desire to the underling platform to scale an Application (not a particular app instance) to a desired count. No mater the present count.  
Still we need to consider how to handle the border case when there are not instances. Because there could be different reasons for it.

tzolov: Also, can we reliably identify a stream app in the target system if we don't know the name of the stream it belongs to?

dturanski: The deployment id is the unique app id which is the app name in cloud foundry, etc. it is unique and dataflow generates the name with the stream name, unique prefix, etc. (group id), etc already.  This API knows nothing about streams.

tzolov: You're right, ignore my comment above. I've wrongly took the deployment for an instance  id. The `deployment id` indeed provides a unique within the stream, versioned application reference.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppDeployer.java
jvalkeal: Yes, leave those tests disabled. We can then require those when time is right.

ilayaperumalg:
LGTM

tzolov:
LGTM too ;)

jvalkeal:
I'll do a quick tests with local deployer and merge then.

PR #324, opened 2019-10-23
Support for eventual app instance count matching

### LINKED ISSUE #323: Eventual app instance count matching
Some deployer operations such as `deploy` and `scale` are non blocking, e.g. the operation will return before the requested (desired) state is reached by the target platform. 

For the scale operation, that means that it may take some time before the target platform deploys the desired number of app instances. 
Therefore we need an instance count matcher that (in combination with the `eventually` matcher) can pull continuously the app instance count until the count matcher is satisfied or the preconfigured timeout threshold fires.

tzolov:
- Add AppCountMatcher class
 - Add AbstractAppDeployerIntegrationTests#appInstanceCount helper method
 - Update doScaleTest to use the eventually(appInstanceCount(is(DESIRED_COUNT))) check.

 Resolves #323


ilayaperumalg:
LGTM, merging.

PR #329, opened 2019-12-13
Move to spring-cloud-dataflow-build

jvalkeal:
- Simple change to use spring-cloud-dataflow-build instead
  of spring-cloud-build.
- Relates spring-cloud/spring-cloud-dataflow#3670

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -12,8 +12,8 @@
 
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
-		<artifactId>spring-cloud-build</artifactId>
-		<version>2.2.0.RELEASE</version>
+		<artifactId>spring-cloud-dataflow-build</artifactId>
+		<version>2.4.0.BUILD-SNAPSHOT</version>
 		<relativePath />
 	</parent>
 


pom.xml
```diff
@@ -12,8 +12,8 @@
 
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
-		<artifactId>spring-cloud-build</artifactId>
-		<version>2.2.0.RELEASE</version>
+		<artifactId>spring-cloud-dataflow-build</artifactId>
```
ilayaperumalg: It looks weird to have the parent artifact Id being `spring-cloud-dataflow-build` for the deployers. Can we come up with any other name that doesn't have `dataflow` in it? Perhaps, we can create `spring-cloud-deployer-build` that acts as a parent for all the deployers?

ilayaperumalg:
I guess we need to create deployer master to be in `2.2.x` across the board for these changes.

pom.xml
jvalkeal: There's going to exactly one parent and there's not good naming for dataflow/skipper/deployers other than they all belong to "dataflow" family.

jvalkeal:
I'm not sure we need to bump deployer versions for this.

ilayaperumalg:
> I'm not sure we need to bump deployer versions for this.

Because the SCDF/Skipper 2.3.x/2.2.x use the same versions of the deployers

jvalkeal:
Right, that's a good point. Completely forget that might become an issue.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as d89354f

PR #333, opened 2020-01-09
Add wagon based http transport

### LINKED ISSUE #331: Allow configuring "preemptive authentication" for remote maven artifactory
As a user, I'm trying to configure "preemptive authentication" against the remote maven artifactory to resolve artifacts with this setting turned-on; however, I'm unable to set that as an option in [`MavenProperties`](https://github.com/spring-cloud/spring-cloud-deployer/blob/master/spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenProperties.java). 

The underlying Maven Resolver library via [`AuthenticationContext`](https://github.com/apache/maven-resolver/blob/master/maven-resolver-api/src/main/java/org/eclipse/aether/repository/AuthenticationContext.java) doesn't appear to support that as an option. Perhaps due to the security concerns described in the official [Maven guide](https://maven.apache.org/guides/mini/guide-http-settings.html#The_Basics).

**Acceptance:**
- Verify whether this is a required setting for `GET` requests (to resolve and download artifacts — maven guide indicates it will help avoid double transfer of creds in HTTP headers _only_ for `PUT` requests
- If it applies to `GET`, study the possibility of adding support to override and to set `usepreemptive` property as "true" (default = "false")
- If the support is not natively available in Maven Resolver, look for alternative add-on or implement support for it in the SPI
- If we end up adding support for it, let's make sure the javadocs and the ref. guide material is updated
- List any limitations and/or security concerns in the docs
- Include test coverage

jvalkeal:
- Order to support preemptive auth need to add wagon http
  as this feature exists only in there.
- Conditionally switch to wagon based http with `maven.use-wagon`
- StaticWagonProvider and StaticWagonConfigurator are a replacement
  for similar classes in a maven resolver core which would only work
  via Plexux containers which are impossible to use in our code base
  as those are mean for real maven dynamic classloading extensions
  with real builds.
- We simply do a static configuration to add `HttpWagon` and its
  configuration which is mapped from `MavenProperties`
- New MavenExtension junit5 concept to mock http endpoints for maven,
  this is something what other tests could use as well.
- Start adding tests for `MavenProperties`.
- Generic list of settings include httpclient params and headers, etc
    wagon.http.all.use-preemptive
    wagon.http.all.use-default-headers
    wagon.http.all.connection-timeout
    wagon.http.all.read-timeout
    wagon.http.all.headers.header1
    wagon.http.all.params.http.connection.timeout
- Method value can be `all`, `put`, `get` or `head` or combination.
- Fixes #331

For an actual dataflow fatjar this causes these additions:
```
98a100
> BOOT-INF/lib/maven-resolver-transport-wagon-1.4.1.jar
218a221,223
> BOOT-INF/lib/wagon-http-3.3.4.jar
> BOOT-INF/lib/wagon-http-shared-3.3.4.jar
> BOOT-INF/lib/wagon-provider-api-3.0.0.jar
```

### DIFF
--- a/spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenArtifactResolver.java
+++ b/spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenArtifactResolver.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2017 the original author or authors.
+ * Copyright 2015-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -39,6 +39,9 @@ import org.eclipse.aether.spi.connector.RepositoryConnectorFactory;
 import org.eclipse.aether.spi.connector.transport.TransporterFactory;
 import org.eclipse.aether.transport.file.FileTransporterFactory;
 import org.eclipse.aether.transport.http.HttpTransporterFactory;
+import org.eclipse.aether.transport.wagon.WagonConfigurator;
+import org.eclipse.aether.transport.wagon.WagonProvider;
+import org.eclipse.aether.transport.wagon.WagonTransporterFactory;
 import org.eclipse.aether.util.artifact.JavaScopes;
 import org.eclipse.aether.util.repository.DefaultProxySelector;
 import org.slf4j.Logger;
@@ -55,6 +58,7 @@ import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 
 /**
  * Resolves a {@link MavenResource} using <a href="https://www.eclipse.org/aether/>aether</a> to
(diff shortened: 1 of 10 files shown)


spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenArtifactResolver.java
```diff
@@ -259,7 +267,15 @@ private RepositorySystem newRepositorySystem() {
 		DefaultServiceLocator locator = MavenRepositorySystemUtils.newServiceLocator();
 		locator.addService(RepositoryConnectorFactory.class, BasicRepositoryConnectorFactory.class);
 		locator.addService(TransporterFactory.class, FileTransporterFactory.class);
-		locator.addService(TransporterFactory.class, HttpTransporterFactory.class);
+
+		if (properties.isUseWagon()) {
+			locator.addService(WagonProvider.class, StaticWagonProvider.class);
```
ilayaperumalg: This is nice as we can add many other providers in the future :-)

ilayaperumalg:
LGTM, verified that the configuration properties are appropriately attached to the MavenArtifactResolver's remote repository system per-repository basis. 

Merging.

ilayaperumalg:
Merged as 2c1306d

Issue #242, opened 2017-11-14
MavenArtifactResolver disregards the  non-proxy-hosts configuration


jvmlet:
Please honor the  `--maven.proxy.non-proxy-hosts` parameter  when building remote repositories.

ilayaperumalg:
Hi @jvmlet , 

The property `maven.proxy.non-proxy-hosts` is already exposed in MavenProperties. Did you mean to say, it is not working as expected?

jvmlet:
Hi
Yes, setting this property has no effect and I don't see any logic  in [MavenArtifactResolver](https://github.com/spring-cloud/spring-cloud-deployer/blob/master/spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenArtifactResolver.java) that honors `non-proxy-hosts`  field from `MavenProperties` object

jvmlet:
Sorry, there is `proxySelector.add(proxy, this.properties.getProxy().getNonProxyHosts());` logic, let me check once again the behaviour...

ilayaperumalg:
@jvmlet Did you get a chance to check this out? Please let us know if you still see the issue.

ilayaperumalg:
Closing this for now and please feel to re-open if you still see any issue.

jvmlet:
Hi @ilayaperumalg , I've checked it . I'm still experience the  `Bad Gateway (502)` problem due to `non-proxy-hosts`  issue.
 I'm running  the local SCDF server with below `java` arguments :

````
-Dhttp.proxyHost=xxxx
-Dhttp.proxyPort=9480
-Dhttp.nonProxyHosts=hotToEscape
-jar spring-cloud-dataflow-server-local-1.3.0.M3.jar
--maven.remote-repositories.repo1.url=http://hotToEscape:8443/artifactory/ASP-Snapshot-Local
--maven.remote-repositories.repo1.auth.username=xxx
--maven.remote-repositories.repo1.auth.password=xxx
--maven.remote-repositories.repo1.snapshot-policy.update-policy=always
--maven.remote-repositories.repo1.snapshot-policy.checksum-policy=warn
--maven.proxy.host=xxx
--maven.proxy.port=9480
--maven.proxy.non-proxy-hosts=hotToEscape
````

jmchen:
This issue is not resolved. Please reopen.
The proxySelector is setup inside DefaultRepositorySystemSession. 
[...]

PR #341, opened 2020-04-26
MavenArtifactResolver disregards the non-proxy-hosts configuration

jmchen:
This issue #242 is not resolved. Please reopen.
The proxySelector is setup inside DefaultRepositorySystemSession. 
However, org.eclipse.aether.transport.http.HttpTransporter is not getting proxy from repositorySession, instead it gets proxy from Repository directly.

`    

    HttpTransporter( RemoteRepository repository, RepositorySystemSession session )
        throws NoTransporterException {

        //... omitted for brevity ...
        proxy = toHost( repository.getProxy() );
        repoAuthContext = AuthenticationContext.forRepository( session, repository );
        proxyAuthContext = AuthenticationContext.forProxy( session, repository );

        state = new LocalState( session, repository, new SslConfig( session, repoAuthContext ) );

        //... omitted for brevity
    }
`

According to document of RepositorySystemSession, the ProxySelector within RepositorySystemSession is not meant for Remote Repository, the repository should have its own Proxy selected.

`

     public interface RepositorySystemSession {
        //... omitted for brevity ...
        /**
         * Gets the proxy selector to use for repositories discovered in artifact descriptors. Note that this selector is
         * not used for remote repositories which are passed as request parameters to the repository system, those
         * repositories are supposed to have their proxy (if any) already set.
         * 
         * @return The proxy selector to use, never {@code null}.
         * @see org.eclipse.aether.repository.RemoteRepository#getProxy()
         * @see RepositorySystem#newResolutionRepositories(RepositorySystemSession, java.util.List)
         */
        ProxySelector getProxySelector();
        //... omitted for brevity ...
    }

`

### DIFF
--- a/spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenArtifactResolver.java
+++ b/spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenArtifactResolver.java
@@ -83,7 +83,7 @@ class MavenArtifactResolver {
 
 	private final List<RemoteRepository> remoteRepositories = new LinkedList<>();
 
-	private final Authentication authentication;
+	private final Authentication proxyAuthentication;
 
 	/**
 	 * Create an instance using the provided properties.
@@ -102,10 +102,10 @@ class MavenArtifactResolver {
 		if (isProxyEnabled() && proxyHasCredentials()) {
 			final String username = this.properties.getProxy().getAuth().getUsername();
 			final String password = this.properties.getProxy().getAuth().getPassword();
-			this.authentication = newAuthentication(username, password);
+			this.proxyAuthentication = newAuthentication(username, password);
 		}
 		else {
-			this.authentication = null;
+			this.proxyAuthentication = null;
 		}
 		File localRepository = new File(this.properties.getLocalRepository());
 		if (!localRepository.exists()) {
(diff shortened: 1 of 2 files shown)


spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenArtifactResolver.java
```diff
@@ -137,28 +137,39 @@ public MavenArtifactResolver(final MavenProperties properties) {
...
+
+				proxy = proxySelector.getProxy(repo);
 			}
-			if (remoteRepositoryHasCredentials(remoteRepository)) {
-				final String username = remoteRepository.getAuth().getUsername();
-				final String password = remoteRepository.getAuth().getPassword();
-				remoteRepositoryBuilder.setAuthentication(newAuthentication(username, password));
-			}
+
+			remoteRepositoryBuilder = new RemoteRepository.Builder(repo);
```
jvalkeal, line 171: Is there a reason why you create a builder again, then set a proxy and build again. Could this be done within `if` block where `proxy` is created?

jmchen: The RemoteRepository is readonly object. It can only be created from RemoteRepository.Builder. DefaultProxySelector only accept RemoteRepository as parameter. In order to utilize ProxySelector, I have to build RemoteRepository first, then call `proxySelector.add(proxy, this.properties.getProxy().getNonProxyHosts());`, then use the `public Builder( RemoteRepository prototype )` to recreate Builder, then build repository again.

The best way is to modify ProxySelector to accept, for example, repository host and protocol as parameter. Unfortunately, the eclipse aether project is already deprecated, it doesn't accept pull request anymore.

jvalkeal: Right that explains it then, thx. Btw eclipse project is defunct but whole project were moved into apache and afaik is still alive. https://github.com/apache/maven-resolver

jmchen: Hey @jvalkeal, thanks for the information. Please let me know if you want the change comes from eclipse aether. I'm not sure the change to eclipse aether would be accepted. It is not really a bug in aether. It would be feature proposal to aether.

jvalkeal:
Squashed/merged per 6a69b3b88b71951eb22ea619d661b012d675b9a5

PR #353, opened 2022-02-10
Add security support to AbstractActuatorTemplate

dturanski:
This can be merged without breaking anything...

### DIFF
--- a/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AbstractActuatorTemplate.java
+++ b/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AbstractActuatorTemplate.java
@@ -16,9 +16,9 @@
 
 package org.springframework.cloud.deployer.spi.app;
 
+import java.nio.charset.Charset;
 import java.util.Collections;
 import java.util.Optional;
-import java.util.function.Consumer;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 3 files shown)


spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AbstractActuatorTemplate.java
```diff
@@ -136,6 +177,16 @@ else if (count > 1) {
 				.findFirst();
 	}
 
-	protected abstract String actuatorUrlForInstance(AppInstanceStatus appInstanceStatus);
+	private Optional<String> prepareDefaultAthentication(AppAdmin appAdmin) {
+		Optional<String> encodeBasicAuth;
```
onobc, line 181: **[TinyNit]** the declaration and assignment can be combined into single statement.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/ActuatorOperations.java
```diff
@@ -29,19 +33,31 @@
...
 	 * @param endpoint the endpoint path relative to the base actuator URL for the instance, with or without preceding '/'.
 	 * @param responseType the expected response type.
+	 * @param requestHeaders optional request headers.
 	 * @return the contents as the given type.
 	 */
-	<T> T getFromActuator(String deploymentId, String guid, String endpoint, Class<T> responseType);
+	<T> T getFromActuator(String deploymentId, String guid, String endpoint, Class<T> responseType,
+			Optional<HttpHeaders> requestHeaders);
+
+	default <T> T getFromActuator (String deploymentId, String guid, String endpoint, Class<T> responseType) {
```
onobc, line 42: I would be in favor of only providing a single method that requires the optional headers. This will not be a primary API where convenience like this would matter. One less API to maintain. 

Whatever we land on, same goes for POST variant.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppAdmin.java
```diff
@@ -0,0 +1,83 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.deployer.spi.app;
+
+import java.util.Map;
+
+import org.springframework.util.StringUtils;
+
+public class AppAdmin {
```
onobc, line 23: A high level description of the purpose of this class may be helpful. It does feel like an internal abstraction as well - if so, might be worth talking about that as well.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AbstractActuatorTemplate.java
```diff
@@ -66,55 +82,80 @@ protected AbstractActuatorTemplate(RestTemplate restTemplate, AppDeployer appDep
...
+
+		requestHeaders.setAccept(Collections.singletonList(MediaType.APPLICATION_JSON));
+		requestHeaders.setContentType(MediaType.APPLICATION_JSON);
+		return requestHeaders;
 	}
 
-	private final <T> ResponseEntity<T> httpGet(String url, Class<T> responseType) {
-		return restTemplate.exchange(url, HttpMethod.GET,
-				new HttpEntity(requestHeaders()), responseType);
+	private void addDefaultAuthorization(HttpHeaders requestHeaders) {
```
onobc, line 142: **[Nit]** Since this is a private 1-liner, may as well inline it to keep the eyes from having to jump down when scanning the code.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AbstractActuatorTemplate.java
```diff
@@ -66,55 +82,80 @@ protected AbstractActuatorTemplate(RestTemplate restTemplate, AppDeployer appDep
...
 	}
 
-	private final String normalizePath(String path) {
-		return path.startsWith("/") ? path : "/" + path;
+
+	private final HttpHeaders requestHeaders(Optional<HttpHeaders> optionalAppInstanceHeaders,
+			Optional<HttpHeaders> optionalRequestHeaders) {
+		HttpHeaders requestHeaders = optionalAppInstanceHeaders
+				.orElse(new HttpHeaders());
+		optionalRequestHeaders.ifPresent(requestHeaders::addAll);
```
onobc, line 130: Do we need to update any tests to get coverage of these header overrides etc? 

Same for the "request headers already contains auth header" scenario.

onobc (commented):
@dturanski looks good - minor comments added.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AbstractActuatorTemplate.java
onobc: Fix it in the merge.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/ActuatorOperations.java
onobc: Let's just leave for now since its a default. Lets copy/pasta the javadocs in the merge.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AppAdmin.java
onobc: Let's add in merge.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/AbstractActuatorTemplate.java
onobc: Adjust it in the merge.

onobc:
Closed via https://github.com/spring-cloud/spring-cloud-deployer/commit/0ea5ca46fa3cd04e88ff39c9ce2557d84ce53557

PR #360, opened 2022-08-25
Adds support for Observations

marcingrzejszczak:
Prerequisites:

- [ ] https://github.com/micrometer-metrics/micrometer/pull/3367
- [ ] https://github.com/micrometer-metrics/tracing/pull/52

Example for Trace View with Zipkin & OTel

image

### DIFF
--- a/spring-cloud-deployer-spi/pom.xml
+++ b/spring-cloud-deployer-spi/pom.xml
@@ -28,8 +28,26 @@
 			<artifactId>reactor-core</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>io.micrometer</groupId>
+			<artifactId>micrometer-observation</artifactId>
+			<version>${micrometer.version}</version>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
 			<groupId>junit</groupId>
 			<artifactId>junit</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>io.micrometer</groupId>
+			<artifactId>micrometer-observation-test</artifactId>
+			<version>${micrometer.version}</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>io.micrometer</groupId>
+			<artifactId>micrometer-tracing-integration-test</artifactId>
+			<version>${micrometer-tracing.version}</version>
 			<scope>test</scope>
 		</dependency>
 	</dependencies>
--- a/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerContext.java
+++ b/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2021 the original author or authors.
+ * Copyright 2013-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -33,9 +33,9 @@ public class AppDeployerContext extends SenderContext<Object> {
 
 	private final RuntimeEnvironmentInfo runtimeEnvironmentInfo;
 
-	private AppDeploymentRequest request;
+	private AppDeploymentRequest deploymentRequest;
 
-	private AppScaleRequest appScaleRequest;
+	private AppScaleRequest scaleRequest;
 
 	private String appId;
 
@@ -47,8 +47,8 @@ public class AppDeployerContext extends SenderContext<Object> {
 	}
 
 	@Nullable
-	public AppDeploymentRequest getRequest() {
-		return request;
+	public AppDeploymentRequest getDeploymentRequest() {
+		return deploymentRequest;
 	}
 
 	@Nullable
@@ -56,17 +56,17 @@ public class AppDeployerContext extends SenderContext<Object> {
 		return runtimeEnvironmentInfo;
 	}
 
-	public void setRequest(AppDeploymentRequest request) {
-		this.request = request;
+	public void setDeploymentRequest(AppDeploymentRequest deploymentRequest) {
+		this.deploymentRequest = deploymentRequest;
 	}
 
 	@Nullable
-	public AppScaleRequest getAppScaleRequest() {
-		return appScaleRequest;
+	public AppScaleRequest getScaleRequest() {
+		return scaleRequest;
 	}
 
-	public void setAppScaleRequest(AppScaleRequest appScaleRequest) {
-		this.appScaleRequest = appScaleRequest;
+	public void setScaleRequest(AppScaleRequest scaleRequest) {
+		this.scaleRequest = scaleRequest;
 	}
 
 	@Nullable
--- a/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerContext.java
+++ b/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerContext.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2013-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.deployer.spi.app.observation;
+
+import io.micrometer.observation.transport.SenderContext;
+
+import org.springframework.cloud.deployer.spi.app.AppScaleRequest;
+import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
+import org.springframework.cloud.deployer.spi.core.RuntimeEnvironmentInfo;
+import org.springframework.lang.Nullable;
+
+// TODO: We don't need to inject tracing headers
+public class AppDeployerContext extends SenderContext<Object> {
+
+	private final RuntimeEnvironmentInfo runtimeEnvironmentInfo;
+
+	private AppDeploymentRequest request;
+
+	private AppScaleRequest appScaleRequest;
+
+	private String appId;
+
+	public AppDeployerContext(RuntimeEnvironmentInfo runtimeEnvironmentInfo) {
+		super((carrier, key, value) -> { });
+		this.runtimeEnvironmentInfo = runtimeEnvironmentInfo;
+		setRemoteServiceName(runtimeEnvironmentInfo.getPlatformType());
+	}
+
+	@Nullable
+	public AppDeploymentRequest getRequest() {
+		return request;
+	}
+
+	@Nullable
+	public RuntimeEnvironmentInfo getRuntimeEnvironmentInfo() {
+		return runtimeEnvironmentInfo;
+	}
+
+	public void setRequest(AppDeploymentRequest request) {
+		this.request = request;
+	}
+
+	@Nullable
+	public AppScaleRequest getAppScaleRequest() {
+		return appScaleRequest;
+	}
+
+	public void setAppScaleRequest(AppScaleRequest appScaleRequest) {
+		this.appScaleRequest = appScaleRequest;
+	}
+
+	@Nullable
+	public String getAppId() {
+		return appId;
+	}
+
+	public void setAppId(String appId) {
+		this.appId = appId;
+	}
+}
--- a/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerContext.java
+++ b/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerContext.java
@@ -18,12 +18,17 @@ package org.springframework.cloud.deployer.spi.app.observation;
 
 import io.micrometer.observation.transport.SenderContext;
 
+import org.springframework.cloud.deployer.spi.app.AppDeployer;
 import org.springframework.cloud.deployer.spi.app.AppScaleRequest;
 import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
 import org.springframework.cloud.deployer.spi.core.RuntimeEnvironmentInfo;
 import org.springframework.lang.Nullable;
 
-// TODO: We don't need to inject tracing headers
+/**
+ * A {@link SenderContext} for {@link AppDeployer}.
+ *
+ * @author Marcin Grzejszczak
+ */
 public class AppDeployerContext extends SenderContext<Object> {
 
 	private final RuntimeEnvironmentInfo runtimeEnvironmentInfo;
@@ -35,7 +40,8 @@ public class AppDeployerContext extends SenderContext<Object> {
 	private String appId;
 
 	public AppDeployerContext(RuntimeEnvironmentInfo runtimeEnvironmentInfo) {
-		super((carrier, key, value) -> { });
+		super((carrier, key, value) -> {
+		});
 		this.runtimeEnvironmentInfo = runtimeEnvironmentInfo;
 		setRemoteServiceName(runtimeEnvironmentInfo.getPlatformType());
 	}
--- a/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerDocumentedObservation.java
+++ b/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerDocumentedObservation.java
@@ -1,0 +1,286 @@
+/*
+ * Copyright 2013-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.deployer.spi.app.observation;
+
+import io.micrometer.common.docs.KeyName;
+import io.micrometer.observation.docs.DocumentedObservation;
+
+enum AppDeployerDocumentedObservation implements DocumentedObservation {
+
+	/**
+	 * Span created upon deploying of an application.
+	 */
+	DEPLOYER_DEPLOY_OBSERVATION {
+		@Override
+		public String getName() {
+			return "spring.cloud.deployer.deploy";
+		}
+
+		@Override
+		public KeyName[] getLowCardinalityKeyNames() {
+			return Tags.values();
+		}
+	},
+
+	/**
+	 * Span created upon undeploying of an application.
+	 */
+	DEPLOYER_UNDEPLOY_OBSERVATION {
+		@Override
+		public String getName() {
+			return "spring.cloud.deployer.undeploy";
+		}
+
+		@Override
+		public KeyName[] getLowCardinalityKeyNames() {
+			return Tags.values();
+		}
+//
+//		@Override
+//		public EventValue[] getEvents() {
+//			return Events.values();
+//		}
+	},
+
+	/**
+	 * Span created upon asking for a status of a deployed application.
+	 */
+	DEPLOYER_STATUS_OBSERVATION {
+		@Override
+		public String getName() {
+			return "spring.cloud.deployer.status";
+		}
+
+		@Override
+		public KeyName[] getLowCardinalityKeyNames() {
+			return Tags.values();
+		}
+//
+//		@Override
+//		public EventValue[] getEvents() {
+//			return Events.values();
+//		}
+	},
+
+	/**
+	 * Span created upon asking for statuses of deployed applications.
+	 */
+	DEPLOYER_STATUSES_OBSERVATION {
+		@Override
+		public String getName() {
+			return "spring.cloud.deployer.statuses";
+		}
+
+		@Override
+		public KeyName[] getLowCardinalityKeyNames() {
+			return Tags.values();
+		}
+
+//		@Override
+//		public EventValue[] getEvents() {
+//			return Events.values();
+//		}
+	},
+
+	/**
+	 * Span created upon asking for logs of deployed applications.
+	 */
+	DEPLOYER_GET_LOG_OBSERVATION {
+		@Override
+		public String getName() {
+			return "spring.cloud.deployer.getLog";
+		}
+
+		@Override
+		public KeyName[] getLowCardinalityKeyNames() {
+			return Tags.values();
+		}
+
+//		@Override
+//		public EventValue[] getEvents() {
+//			return Events.values();
+//		}
+	},
+
+	/**
+	 * Span created upon asking for logs of deployed applications.
+	 */
+	DEPLOYER_SCALE_SPAN {
+		@Override
+		public String getName() {
+			return "spring.cloud.deployer.scale";
+		}
+
+		@Override
+		public KeyName[] getLowCardinalityKeyNames() {
+			return KeyName.merge(Tags.values(), ScaleTags.values());
+		}
+//
+//		@Override
+//		public EventValue[] getEvents() {
+//			return Events.values();
+//		}
+	};
+
+	enum Tags implements KeyName {
+
+		/**
+		 * Name of the platform to which apps are being deployed.
+		 */
+		PLATFORM_NAME {
+			@Override
+			public String asString() {
+				return "spring.cloud.deployer.platform.name";
+			}
+		},
+
+		/**
+		 * ID of the deployed application.
+		 */
+		APP_ID {
+			@Override
+			public String asString() {
+				return "spring.cloud.deployer.app.id";
+			}
+		},
+
+		/**
+		 * Name of the deployed application.
+		 */
+		APP_NAME {
+			@Override
+			public String asString() {
+				return "spring.cloud.deployer.app.name";
+			}
+		},
+
+		/**
+		 * Group of the deployed application.
+		 */
+		APP_GROUP {
+			@Override
+			public String asString() {
+				return "spring.cloud.deployer.app.group";
+			}
+		},
+
+		/**
+		 * CloudFoundry API URL.
+		 */
+		CF_URL {
+			@Override
+			public String asString() {
+				return "spring.cloud.deployer.platform.cf.url";
+			}
+		},
+
+		/**
+		 * CloudFoundry org.
+		 */
+		CF_ORG {
+			@Override
+			public String asString() {
+				return "spring.cloud.deployer.platform.cf.org";
+			}
+		},
+
+		/**
+		 * CloudFoundry space.
+		 */
+		CF_SPACE {
+			@Override
+			public String asString() {
+				return "spring.cloud.deployer.platform.cf.space";
+			}
+		},
+
+		/**
+		 * Kubernetes API URL.
+		 */
+		K8S_URL {
+			@Override
+			public String asString() {
+				return "spring.cloud.deployer.platform.k8s.url";
+			}
+		},
+
+		/**
+		 * Kubernetes namespace.
+		 */
+		K8S_NAMESPACE {
+			@Override
+			public String asString() {
+				return "spring.cloud.deployer.platform.k8s.namespace";
+			}
+		},
+
+	}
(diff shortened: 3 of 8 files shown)


spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/DefaultAppDeployerObservationConvention.java
```diff
@@ -0,0 +1,94 @@
...
+import org.springframework.util.StringUtils;
+
+public class DefaultAppDeployerObservationConvention implements AppDeployerObservationConvention {
+
+	// TODO: Check what is high and what is low cardinality
+	@Override
+	public KeyValues getLowCardinalityKeyValues(AppDeployerContext context) {
+		KeyValues keyValues = KeyValues.empty();
+		if (context.getAppId() != null) {
+			keyValues = keyValues.and(AppDeployerDocumentedObservation.Tags.APP_ID.withValue(context.getAppId()));
```
jonatan-ivanov, line 35: Static imports might help increasing the readability of these statements.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/DefaultAppDeployerObservationConvention.java
```diff
@@ -0,0 +1,94 @@
...
+			String group = request.getDeploymentProperties().get("spring.cloud.deployer.group");
+			if (StringUtils.hasText(group)) {
+				keyValues = keyValues.and(AppDeployerDocumentedObservation.Tags.APP_GROUP.withValue(group));
+			}
+		}
+		keyValues = keyValues.and(addCfTags(platformSpecificInfo));
+		return keyValues.and(addK8sTags(platformSpecificInfo));
+	}
+
+	private KeyValues addCfTags(Map<String, String> platformSpecificInfo) {
```
jonatan-ivanov, line 69: Should not this be `getCfTags`?

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/DefaultAppDeployerObservationConvention.java
```diff
@@ -0,0 +1,94 @@
...
+		if (platformSpecificInfo.containsKey("Organization")) {
+			keyValues = keyValues.and(AppDeployerDocumentedObservation.Tags.CF_ORG.withValue(platformSpecificInfo.get("Organization")));
+		}
+		if (platformSpecificInfo.containsKey("Space")) {
+			keyValues = keyValues.and(AppDeployerDocumentedObservation.Tags.CF_SPACE.withValue(platformSpecificInfo.get("Space")));
+		}
+		return keyValues;
+	}
+
+	private KeyValues addK8sTags(Map<String, String> platformSpecificInfo) {
```
jonatan-ivanov, line 83: Should not this be `getK8sTags`?

spring-cloud-deployer-spi/src/test/java/org/springframework/cloud/deployer/spi/app/observation/ObservedAppDeployerTests.java
```diff
@@ -0,0 +1,144 @@
...
+
+			String id = appDeployer.deploy(deploymentRequest());
+			appDeployer.status(id);
+			appDeployer.scale(new AppScaleRequest(id, 2));
+			appDeployer.undeploy(id);
+
+			SpansAssert.assertThat(bb.getFinishedSpans())
+					.haveSameTraceId()
+					.hasASpanWithName("deploy", spanAssert -> spanAssert.hasRemoteServiceNameEqualTo("Test")
+							.hasTag("spring.cloud.deployer.app.id", "Deployment request received")
```
jonatan-ivanov, line 60: Is this what the app id should be?

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerDocumentedObservation.java
```diff
@@ -0,0 +1,353 @@
...
+	 */
+	DEPLOYER_GET_LOG_OBSERVATION {
+		@Override
+		public Class<? extends Observation.ObservationConvention<? extends Observation.Context>> getDefaultConvention() {
+			return DefaultAppDeployerObservationConvention.class;
+		}
+
+		@Override
+		public String getName() {
+			return "spring.cloud.deployer.getLog";
```
cppwfs, line 158: Curious as to why the name and contextual name is `getLog` versus `log`.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerContext.java
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright 2013-2021 the original author or authors.
```
cppwfs, line 2: Copyrights need to be updated to 2022.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerDocumentedObservation.java
```diff
@@ -0,0 +1,353 @@
...
+			@Override
+			public String asString() {
+				return "spring.cloud.deployer.platform.k8s.url";
+			}
+		},
+
+		/**
+		 * Kubernetes namespace.
+		 */
+		K8S_NAMESPACE {
```
cppwfs, line 292: Thinking ... I wonder if we have platform specific Tags for the local deployer?

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/ObservedAppDeployer.java
```diff
@@ -0,0 +1,227 @@
...
+		this.customConvention = customConvention;
+		this.pollDelay = pollDelay;
+	}
+
+	public ObservedAppDeployer(ObservationRegistry observationRegistry, AppDeployer delegate, Long pollDelay) {
+		this(observationRegistry, null, delegate, pollDelay);
+	}
+
+	@Override
+	public String deploy(AppDeploymentRequest request) {
```
cppwfs, line 67: We also have a `TaskLauncher.launch` and `TaskLauncher.cancel` for ephemeral applications (Jobs).

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/ObservedAppDeployer.java
```diff
@@ -0,0 +1,227 @@
...
+		AppDeployerContext context = new AppDeployerContext(environmentInfo());
```
onobc, line 68: I think a fluent builder on ACD may work out nice in here. Something like 
```java
AppDeployerContext context = AppDeployerContext.for(environmentInfo()).withDeployRequest(request);
```

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerContext.java
```diff
@@ -0,0 +1,80 @@
...
+import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
+import org.springframework.cloud.deployer.spi.core.RuntimeEnvironmentInfo;
+import org.springframework.lang.Nullable;
+
+/**
+ * A {@link SenderContext} for {@link AppDeployer}.
+ *
+ * @author Marcin Grzejszczak
+ */
+public class AppDeployerContext extends SenderContext<Object> {
```
onobc, line 32: Where can I find out more about the carrier (aka `C`) in `SenderContext<C>`? I am curious why its `Object` here.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerDocumentedObservation.java
```diff
@@ -0,0 +1,353 @@
...
+			return KeyName.merge(Tags.values(), ScaleTags.values());
+		}
+
+		@Override
+		public Event[] getEvents() {
+			return Events.values();
+		}
+	};
+
+	enum Tags implements KeyName {
```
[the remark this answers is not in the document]
onobc: ➕ 1

I think tags is an overloaded term and that `<thing>KeyName` is more telling. I think these are `DeployKeyName`s though as they are centered around the un/deploy operations.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerDocumentedObservation.java
```diff
@@ -0,0 +1,353 @@
...
+				return "spring.cloud.deployer.scale.count";
+			}
+		}
+
+	}
+
+	enum Events implements Event {
+
+		/**
+		 * When deployer started deploying the application.
```
onobc, line 328: This is used also by undeploy. Should we have another one for undeploy or just update the docs here?

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerDocumentedObservation.java
```diff
@@ -0,0 +1,353 @@
...
+import io.micrometer.observation.Observation;
+import io.micrometer.observation.Observation.Event;
+import io.micrometer.observation.docs.DocumentedObservation;
+
+enum AppDeployerDocumentedObservation implements DocumentedObservation {
+
+	/**
+	 * Observation created upon deploying of an application.
+	 */
+	DEPLOYER_DEPLOY_OBSERVATION {
```
onobc, line 29: The prefix of `DEPLOYER` is a bit redundant as the `AppDeployerDocumentedObservation.DEPLOY_OBSERVATION` implies that.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/DefaultAppDeployerObservationConvention.java
```diff
@@ -0,0 +1,94 @@
...
+import io.micrometer.common.KeyValues;
+
+import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
+import org.springframework.cloud.deployer.spi.core.RuntimeEnvironmentInfo;
+import org.springframework.lang.Nullable;
+import org.springframework.util.StringUtils;
+
+public class DefaultAppDeployerObservationConvention implements AppDeployerObservationConvention {
+
+	// TODO: Check what is high and what is low cardinality
```
onobc, line 30: Good idea. What threshold is considered low vs. high cardinality these days?

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/DefaultAppDeployerObservationConvention.java
```diff
@@ -0,0 +1,94 @@
...
+				keyValues = keyValues.and(AppDeployerDocumentedObservation.Tags.APP_GROUP.withValue(group));
+			}
+		}
+		keyValues = keyValues.and(addCfTags(platformSpecificInfo));
+		return keyValues.and(addK8sTags(platformSpecificInfo));
+	}
+
+	private KeyValues addCfTags(Map<String, String> platformSpecificInfo) {
+		KeyValues keyValues = KeyValues.empty();
+		if (platformSpecificInfo.containsKey("API Endpoint")) {
```
onobc, line 71: Where do these string constants come from? Same `?` for the K8S ones below.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/ObservedAppDeployer.java
```diff
@@ -0,0 +1,227 @@
...
+			else if (this.current == null) {
+				throw new IllegalStateException("Current state can't be null");
+			}
+			DeploymentState currentState = this.current.getState();
+			DeploymentState previousState = this.previous.getState();
+			return currentState != previousState;
+		}
+
+		private boolean isFinished() {
+			boolean finished = this.current.getState() == DeploymentState.deployed
```
onobc, line 214: May be simpler to invert this where it is `NOT (deploying OR partial)`

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/ObservedAppDeployer.java
```diff
@@ -0,0 +1,227 @@
...
+		AppDeployerContext context = new AppDeployerContext(environmentInfo());
+		context.setAppId(id);
+		Observation observation = AppDeployerDocumentedObservation.DEPLOYER_STATUS_OBSERVATION.start(this.customConvention, DEFAULT_CONVENTION, context, this.observationRegistry);
+		return observation.observe(() -> this.delegate.status(id));
+	}
+
+	@Override
+	public Mono<AppStatus> statusReactive(String id) {
+		return this.delegate.statusReactive(id);
+		// TODO: Fix me
```
onobc, line 119: What is the story w/ these droids?

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/ObservedAppDeployer.java
```diff
@@ -0,0 +1,227 @@
...
+			if (log.isDebugEnabled()) {
+				log.debug("Current observation is [" + observation + "]");
+			}
+		}
+
+		private PreviousAndCurrentStatus updateCurrent(AppStatus current) {
+			if (log.isTraceEnabled()) {
+				log.trace("State before change: current [" + this.current + "], previous [" + this.previous + "]");
+			}
+			this.previous = this.current;
```
onobc, line 176: Any thread-safety concerns here? If so, and we only care about the previous to register a status changed or not, we could leverage AtomicReference using AppStatus as the ref.

spring-cloud-deployer-spi/src/test/java/org/springframework/cloud/deployer/spi/app/observation/ObservedAppDeployerTests.java
```diff
@@ -0,0 +1,144 @@
...
+import org.springframework.cloud.deployer.spi.app.AppDeployer;
+import org.springframework.cloud.deployer.spi.app.AppScaleRequest;
+import org.springframework.cloud.deployer.spi.app.AppStatus;
+import org.springframework.cloud.deployer.spi.core.AppDefinition;
+import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
+import org.springframework.cloud.deployer.spi.core.RuntimeEnvironmentInfo;
+import org.springframework.core.io.PathResource;
+
+/**
+ * Tests for {@link AppDeployer}
```
onobc, line 39: **[TinyNit]** ObservedAppDeployer

spring-cloud-deployer-spi/src/test/java/org/springframework/cloud/deployer/spi/app/observation/ObservedAppDeployerTests.java
```diff
@@ -0,0 +1,144 @@
...
+ */
+public class ObservedAppDeployerTests extends SampleTestRunner {
```
onobc, line 41: Oooh `SampleTestRunner` - I like this!

spring-cloud-deployer-spi/src/test/java/org/springframework/cloud/deployer/spi/app/observation/ObservedAppDeployerTests.java
```diff
@@ -0,0 +1,144 @@
...
+							.hasTag("spring.cloud.deployer.scale.count", "2")
+							.hasKindEqualTo(Span.Kind.PRODUCER))
+					.hasASpanWithName("undeploy", spanAssert -> spanAssert.hasRemoteServiceNameEqualTo("Test")
+							.hasTag("spring.cloud.deployer.app.id", "Deployment request received")
+							.hasTag("spring.cloud.deployer.platform.k8s.url", "https://foo.bar/baz")
+							.hasKindEqualTo(Span.Kind.PRODUCER))
+					.hasSize(4);
+
+			MeterRegistryAssert.assertThat(getMeterRegistry())
+					.hasTimerWithNameAndTags("spring.cloud.deployer.deploy", Tags.of("error", "none", "spring.cloud.deployer.app.id", "Deployment request received", "spring.cloud.deployer.platform.k8s.url", "https://foo.bar/baz"))
```
onobc, line 78: I would be nice to have at least 1 error case where maybe deploy/scale throws an exception.

onobc (changes requested):
@marcingrzejszczak thanks for adding this. It is great to see it.

Most of my comments are of the nit variety. Take what you like - leave the rest ;) 

I do have a bigger question though, since `AppDeployer` is an interface and the current impls already extend an ABC what is the plan for "hooking this in" ? 

```java
public class KubernetesAppDeployer extends AbstractKubernetesDeployer implements AppDeployer {


public class CloudFoundryAppDeployer extends AbstractCloudFoundryDeployer implements MultiStateAppDeployer 
```

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerDocumentedObservation.java
marcingrzejszczak: It can be `log` - the method name is `getLog` so I wanted to be consistent.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerContext.java
```diff
@@ -0,0 +1,80 @@
...
+	private final RuntimeEnvironmentInfo runtimeEnvironmentInfo;
+
+	private AppDeploymentRequest request;
+
+	private AppScaleRequest appScaleRequest;
+
+	private String appId;
+
+	public AppDeployerContext(RuntimeEnvironmentInfo runtimeEnvironmentInfo) {
+		super((carrier, key, value) -> {
```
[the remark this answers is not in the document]
marcingrzejszczak: In vast majority of cases you should provide the way to instrument the carrier. In this case we don't need to instrument the carrier and still want to see the `producer` / `consumer` flow in Zipkin

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerDocumentedObservation.java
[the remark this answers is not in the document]
marcingrzejszczak: I think these are platform related. I'd go with `PlatformKeyNames` or sth like that

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/DefaultAppDeployerObservationConvention.java
marcingrzejszczak: I'll change it to just `cfTags`

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerContext.java
marcingrzejszczak: That's because we don't have any real carrier here. We're using this approach with `SenderContext` and the corresponding tracing handlers to set the span kinds and remote service names automatically

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerContext.java
```diff
@@ -0,0 +1,80 @@
...
+		});
+		this.runtimeEnvironmentInfo = runtimeEnvironmentInfo;
+		setRemoteServiceName(runtimeEnvironmentInfo.getPlatformType());
+	}
+
+	@Nullable
```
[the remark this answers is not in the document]
marcingrzejszczak: What do you mean?

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerDocumentedObservation.java
marcingrzejszczak: I'm updating the docs

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/DefaultAppDeployerObservationConvention.java
marcingrzejszczak: Jonatan has created a `HighCardinalityTagsDetector` class that uses the following logic to calculate that an entry goes from low to high cardinality

```java
private static long calculateThreshold() {
        // 10% of the heap in MiB
        long allowance = Runtime.getRuntime().maxMemory() / 1024 / 1024 / 10;

        // 2k Meters can take ~1MiB, 2M Meters can take ~1GiB
        return Math.max(1_000, Math.min(allowance * 2_000, 2_000_000));
    }
```

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/DefaultAppDeployerObservationConvention.java
marcingrzejszczak: I think I looked at the Kubernetes and CF deployers. So this instrumentation is taken from Sleuth where I had to assume everything there. Now with Spring Observability what we can do is move the platform specific code to the corresponding deployers. Then all the platform specific tagging would be removed from this class. WDYT?

cppwfs: I agree.  This also covers my comment about Local deployer as well as the condition when we add a new deployer for another platform.

marcingrzejszczak: I'll leave it for now here so that we agree to the principles and then in `kubernetes` or `cf` deployers we will just add global conventions that will set the key values

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/ObservedAppDeployer.java
```diff
@@ -0,0 +1,227 @@
...
+		PreviousAndCurrentStatus previousAndCurrentStatus = new PreviousAndCurrentStatus(observation);
+		// @formatter:off
+			this.delegate.statusReactive(id)
+					.map(previousAndCurrentStatus::updateCurrent)
+					.repeatWhen(repeat -> repeat.flatMap(i -> Mono.delay(Duration.ofMillis(this.pollDelay))))
+					.takeUntil(PreviousAndCurrentStatus::isFinished)
+					.last()
+					.doOnNext(PreviousAndCurrentStatus::eventOnObservation)
+					.doOnError(observation::error)
+					// we will close the span in the reactive part
```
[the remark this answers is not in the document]
marcingrzejszczak: actually this reactive part ;) let me remove this

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/ObservedAppDeployer.java
marcingrzejszczak: I'll fix this

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/ObservedAppDeployer.java
marcingrzejszczak: I don't think so - here we have statuses changing for a single id.

spring-cloud-deployer-spi/src/test/java/org/springframework/cloud/deployer/spi/app/observation/ObservedAppDeployerTests.java
marcingrzejszczak: That's what I put in here https://github.com/spring-cloud/spring-cloud-deployer/pull/360/files/f0b16ccfb96ca1e1e68be28f163944fa931b206a#diff-bc19637713d0dd0728458d6c4b297b715590a3d59243646e727e700d02efd16bR106 :shrug:

spring-cloud-deployer-spi/pom.xml
```diff
@@ -27,11 +27,29 @@
 			<groupId>io.projectreactor</groupId>
 			<artifactId>reactor-core</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>io.micrometer</groupId>
+			<artifactId>micrometer-observation</artifactId>
+			<version>${micrometer.version}</version>
```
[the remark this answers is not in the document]
marcingrzejszczak: I need to override this cause Boot comes with `-M4` AFAIR and we need to use snapshots

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerContext.java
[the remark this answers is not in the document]
onobc: Makes sense. What exactly is the "carrier"? Can you point me to where I can learn more about this? Thx.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerContext.java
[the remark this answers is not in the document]
onobc: Its a total nit, I meant the get/set not adjacent to one another, here I think its just because getRuntimeEI is plopped in between the get/setDeploymentRequest

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/DefaultAppDeployerObservationConvention.java
onobc: Sounds great. Should we add a TODO in here, create an issue for this, OR just sit back and relax and know that @marcingrzejszczak will made this change? ;). I'm fine w/ any of these reminder mechanisms.

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/ObservedAppDeployer.java
```diff
@@ -0,0 +1,232 @@
...
+
+		private final Observation observation;
+
+		private AppStatus current;
+
+		private AppStatus previous;
+
+		private PreviousAndCurrentStatus(Observation observation) {
+			this.observation = observation;
+			if (log.isDebugEnabled()) {
```
onobc, line 176: Thank you for adding `LogAccessor`. Do you mind updating these calls to take advantage of it? These become 
```java
log.debug(() -> "Current observation is [" + observation + "]";
```
We get to drop the pesky "is<level>Enabled" as the internal impl does that and only calls supplier when its enabled.

onobc (commented):
@marcingrzejszczak thanks for all of the updates. Looks good to me. I only have 2 more asks:
* using the log accessor 
* the bigger question I asked [here](https://github.com/spring-cloud/spring-cloud-deployer/pull/360#pullrequestreview-1089545065)

spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/app/observation/AppDeployerDocumentedObservation.java
```diff
@@ -0,0 +1,353 @@
...
+	DEPLOY_OBSERVATION {
+		@Override
+		public Class<? extends Observation.ObservationConvention<? extends Observation.Context>> getDefaultConvention() {
+			return DefaultAppDeployerObservationConvention.class;
+		}
+
+		@Override
+		public String getName() {
+			return "spring.cloud.deployer.deploy";
+		}
```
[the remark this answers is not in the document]
marcingrzejszczak: We should fix that then cause otherwise we would need 4 different conventions or if / else statement in the convention which is impossible to parse

marcingrzejszczak:
Closing this issue after talking to the SCDF team

PR #381, opened 2023-04-13
Respect initContainer from global properties

onobc:
- The init container coming from the KubernetesDeployerProperties (ie from the platform account props) was not being respected

Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/5280


spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DeploymentPropertiesResolver.java
```diff
@@ -584,50 +584,53 @@ Affinity getAffinityRules(Map<String, String> kubernetesDeployerProperties) {
...
 				this.propertyPrefix + ".initContainer", "initContainer");
 
-		if (deployerProperties.getInitContainer() == null) {
-			String containerName = PropertyParserUtils.getDeploymentPropertyValue(kubernetesDeployerProperties,
-					this.propertyPrefix + ".initContainer.containerName");
-
-			String imageName = PropertyParserUtils.getDeploymentPropertyValue(kubernetesDeployerProperties,
-					this.propertyPrefix + ".initContainer.imageName");
-
-			String commands = PropertyParserUtils.getDeploymentPropertyValue(kubernetesDeployerProperties,
```
onobc, line 598: **[NOTE]** This was a bug lurking in the shadows.. It was not parsing the csv command string and just passing into the single string builder method. It is fixed now to do so.

PR #385, opened 2023-04-20
[CI] Add trivy scan for PRs

onobc:


corneil:
I missed this and implemented the jobs.

Issue #386, opened 2023-04-20
Is there any deployer properties with 'terminationGracePeriodSeconds'


Kim-menu:

onobc:
**NOTE** Since k8s 1.25, there is support for a probe-level `terminationGracePeriodSeconds`. That one is a bit more involved than this simple property as there are N probes to handle - not complicated, just in multiple places to update and test. Let's see if the pod-level setting suffices for now.

juanpablo-santos:
Hi, 

minor nitpick on this: on SCDF's launch task screen, both in Builder and Freetext forms, this new property is translated with a typo, `deployer.<task>.terminiation-grace-period-seconds`. This misspelling appears to come from [`KubernetesDeployerProperties`](https://github.com/spring-cloud/spring-cloud-deployer/blob/main/spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java#L64). Defining this property at platform level also requires to misspell it.

Other than that, super useful feature, thanks!

onobc:
Oh my, sorry about this @juanpablo-santos . It is more than a nitpick IMO. 
We will get this fixed and be sure the tests in place are paying more attention to the naming of the field. 

In the meantime, here is the workaround re-stated:
- when specifying the property at platform level use the bad spelling `terminiation-grace-period-seconds`
- when specifying the property at app or deployer level use the proper spelling `termination-grace-period-seconds`

onobc:
Closing in favor of #394

PR #387, opened 2023-04-20
Add pod-level terminationGracePeriodSeconds k8s deployer prop

onobc:
See #386


onobc:
cc: @Kim-menu

spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
```diff
@@ -58,6 +58,11 @@ public class KubernetesDeployerProperties {
      */
     private int maximumConcurrentTasks = 20;
 
+	/**
```
onobc, line 61: These files all need fully reformatting - I will do that in a separate PR so that we can see the forest through the trees in this PR.

PR #395, opened 2023-05-18
Correct spelling for terminationGracePeriodSeconds

### LINKED ISSUE #394: Correct spelling of terminationGracePeriodSecs property
The new property added in #386 was misspelled. This ticket is to correct that. 

Here is the workaround to successfully use the property:
- when specifying the property at platform level use the bad spelling `terminiation-grace-period-seconds`
- when specifying the property at app or deployer level use the proper spelling `termination-grace-period-seconds`

cppwfs:
resolves #394


onobc (approved):
LGTM! Thanks @cppwfs ✋🏻

PR #404, opened 2023-09-14
Add random number to k8s task deployment id.

### LINKED ISSUE #399: Duplicate Kubernetes pod ids when tasks are launched at the same time
When the Spring Cloud DataFlow Server in the Kubernetes environment tries to launch pods for many tasks(200+), we have encountered scenarios where duplicate Pod id is generated for 3-4 pods. I have looked at the [code generating the id](https://github.com/spring-cloud/spring-cloud-deployer/blob/00f9473b790b117ba0fce87dfcc501bf1de0ac0e/spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java#L228), and it encodes the timestamp in milliseconds using the hashid library.

Multiple tasks end up launched simultaneously and with the same pod id. Can we introduce randomness here or another way to ensure pod id uniqueness is not dependent on only the timestamp?

onobc:
Avoid duplicate deployment id when multiple tasks launched simultaneously.

Fixes #399

### DIFF
--- a/spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java
+++ b/spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java
@@ -227,7 +227,7 @@ public class KubernetesTaskLauncher extends AbstractKubernetesDeployer implement
 	protected String createDeploymentId(AppDeploymentRequest request) {
 		String name = request.getDefinition().getName();
 		Hashids hashids = new Hashids(name, 0, "abcdefghijklmnopqrstuvwxyz1234567890");
-		long idToEncode = System.currentTimeMillis() + ThreadLocalRandom.current().nextInt(100);
+		long idToEncode = System.currentTimeMillis() + ThreadLocalRandom.current().nextInt(1000);
 		String hashid = hashids.encode(idToEncode);
 		String deploymentId = name + "-" + hashid;
 		// Kubernetes does not allow . in the name and does not allow uppercase in the name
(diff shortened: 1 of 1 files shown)


spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java
corneil, line 1: We can use currentNanos plus a randon amount or a larger random amount added to the millis.

spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java
```diff
@@ -227,7 +227,7 @@ private boolean maxConcurrentExecutionsReached() {
 	protected String createDeploymentId(AppDeploymentRequest request) {
 		String name = request.getDefinition().getName();
 		Hashids hashids = new Hashids(name, 0, "abcdefghijklmnopqrstuvwxyz1234567890");
-		long idToEncode = System.currentTimeMillis() + ThreadLocalRandom.current().nextInt(100);
+		long idToEncode = System.nanoTime() + ThreadLocalRandom.current().nextInt(100);
```
onobc, line 230: There seems to be not much reward for doing this. The only guarantee is that it is at least as good as System.currentTimeMillis(). Also it mentions that at some point in the future the numbers may be negative. Also, its late in the release cycle and I would prefer not to bring in another unknown.

```java
     * This method can only be used to measure elapsed time and is
     * not related to any other notion of system or wall-clock time.
     * The value returned represents nanoseconds since some fixed but
     * arbitrary <i>origin</i> time (perhaps in the future, so values
     * may be negative).  The same origin is used by all invocations of
     * this method in an instance of a Java virtual machine; other
     * virtual machine instances are likely to use a different origin.
     *
     * <p>This method provides nanosecond precision, but not necessarily
     * nanosecond resolution (that is, how frequently the value changes)
     * - no guarantees are made except that the resolution is at least as
     * good as that of {@link #currentTimeMillis()}.
```

corneil: Then we have to increase the range of random numbers or have an AtomicLong as the number.

PR #413, opened 2023-12-04
Retrieve CF log for tasks and  streams

cppwfs:
* Filters null envelopes from results.  This was causing the original failure from our update to use the `LogCacheClient`.
* Retrieve GUID from app instance for stream apps. 
* Update `AbstractCloudFoundryTaskLauncher` to add the `app-cf-guid` to the attributes for a task.
* Updated CloudFoundryTaskLauncher to use the `app-cf-guid` from the attributes and use this to retrieve logs from the `LogCacheClient`
* Add Tests

### DIFF
--- a/spring-cloud-deployer-cloudfoundry/src/main/java/org/springframework/cloud/deployer/spi/cloudfoundry/AbstractCloudFoundryTaskLauncher.java
+++ b/spring-cloud-deployer-cloudfoundry/src/main/java/org/springframework/cloud/deployer/spi/cloudfoundry/AbstractCloudFoundryTaskLauncher.java
@@ -149,6 +149,9 @@ abstract class AbstractCloudFoundryTaskLauncher extends AbstractCloudFoundryDepl
 	protected TaskStatus toTaskStatus(GetTaskResponse response) {
 		Map<String, String> attributes = new HashMap<>();
 		//retrieve CF-GUID for retrieving logs
+		Assert.notNull(response.getTaskRelationships(), "response must contain task relationships.");
+		Assert.notNull(response.getTaskRelationships().getApp(), "app in the taskRelationships of the response must not be null");
+		Assert.notNull(response.getTaskRelationships().getApp().getData(), "data in the app of the task relationships within the response must not be null");
 		attributes.put("app-cf-guid", response.getTaskRelationships().getApp().getData().getId());
 		switch (response.getState()) {
 		case SUCCEEDED:
--- a/spring-cloud-deployer-cloudfoundry/src/main/java/org/springframework/cloud/deployer/spi/cloudfoundry/AbstractCloudFoundryTaskLauncher.java
+++ b/spring-cloud-deployer-cloudfoundry/src/main/java/org/springframework/cloud/deployer/spi/cloudfoundry/AbstractCloudFoundryTaskLauncher.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2019 the original author or authors.
+ * Copyright 2016-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,6 +17,8 @@
 package org.springframework.cloud.deployer.spi.cloudfoundry;
 
 import java.time.Duration;
+import java.util.HashMap;
+import java.util.Map;
 
 import io.jsonwebtoken.lang.Assert;
 import org.cloudfoundry.client.CloudFoundryClient;
@@ -145,17 +147,20 @@ abstract class AbstractCloudFoundryTaskLauncher extends AbstractCloudFoundryDepl
 	}
 
 	protected TaskStatus toTaskStatus(GetTaskResponse response) {
+		Map<String, String> attributes = new HashMap<>();
+		//retrieve CF-GUID for retrieving logs
+		attributes.put("app-cf-guid", response.getTaskRelationships().getApp().getData().getId());
 		switch (response.getState()) {
 		case SUCCEEDED:
-			return new TaskStatus(response.getId(), LaunchState.complete, null);
+			return new TaskStatus(response.getId(), LaunchState.complete, attributes);
 		case RUNNING:
-			return new TaskStatus(response.getId(), LaunchState.running, null);
+			return new TaskStatus(response.getId(), LaunchState.running, attributes);
 		case PENDING:
-			return new TaskStatus(response.getId(), LaunchState.launching, null);
+			return new TaskStatus(response.getId(), LaunchState.launching, attributes);
 		case CANCELING:
-			return new TaskStatus(response.getId(), LaunchState.cancelled, null);
+			return new TaskStatus(response.getId(), LaunchState.cancelled, attributes);
 		case FAILED:
-			return new TaskStatus(response.getId(), LaunchState.failed, null);
+			return new TaskStatus(response.getId(), LaunchState.failed, attributes);
 		default:
 			throw new IllegalStateException(String.format("Unsupported CF task state %s", response.getState()));
 		}
--- a/spring-cloud-deployer-cloudfoundry/src/main/java/org/springframework/cloud/deployer/spi/cloudfoundry/ApplicationLogAccessor.java
+++ b/spring-cloud-deployer-cloudfoundry/src/main/java/org/springframework/cloud/deployer/spi/cloudfoundry/ApplicationLogAccessor.java
@@ -56,12 +56,11 @@ public class ApplicationLogAccessor {
      * @return String containing the log information or empty string if no entries are available.
      */
     public String getLog(String deploymentId, Duration apiTimeout) {
-        logger.info("getLog:{}:{}", deploymentId, apiTimeout);
+        logger.debug("Retrieving log for deploymentId:{} with apiTimeout:{}", deploymentId, apiTimeout);
         Assert.hasText(deploymentId, "id must have text and not null");
         Assert.notNull(apiTimeout, "apiTimeout must not be null");
         StringBuilder stringBuilder = new StringBuilder();
         ReadRequest request = ReadRequest.builder().sourceId(deploymentId).build();
-        Assert.notNull(request, "request must not be null");
         List<Log> logs = this.logCacheClient
                 .read(request)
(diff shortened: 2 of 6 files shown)


spring-cloud-deployer-cloudfoundry/src/main/java/org/springframework/cloud/deployer/spi/cloudfoundry/ApplicationLogAccessor.java
```diff
@@ -54,29 +56,32 @@ public ApplicationLogAccessor(LogCacheClient logCacheClient) {
      * @return String containing the log information or empty string if no entries are available.
      */
     public String getLog(String deploymentId, Duration apiTimeout) {
+        logger.info("getLog:{}:{}", deploymentId, apiTimeout);
         Assert.hasText(deploymentId, "id must have text and not null");
         Assert.notNull(apiTimeout, "apiTimeout must not be null");
         StringBuilder stringBuilder = new StringBuilder();
+        ReadRequest request = ReadRequest.builder().sourceId(deploymentId).build();
+        Assert.notNull(request, "request must not be null");
```
onobc, line 64: I don't think this can ever be null since we build it ourselves in the previous line.

spring-cloud-deployer-cloudfoundry/src/main/java/org/springframework/cloud/deployer/spi/cloudfoundry/ApplicationLogAccessor.java
```diff
@@ -54,29 +56,32 @@ public ApplicationLogAccessor(LogCacheClient logCacheClient) {
      * @return String containing the log information or empty string if no entries are available.
      */
     public String getLog(String deploymentId, Duration apiTimeout) {
+        logger.info("getLog:{}:{}", deploymentId, apiTimeout);
```
onobc, line 59: Let's consider moving this to debug as it is really only interesting when debugging an issue. IMO

spring-cloud-deployer-cloudfoundry/src/main/java/org/springframework/cloud/deployer/spi/cloudfoundry/ApplicationLogAccessor.java
```diff
@@ -54,29 +56,32 @@ public ApplicationLogAccessor(LogCacheClient logCacheClient) {
...
         });
+
         return stringBuilder.toString();
     }
 
     private Flux<Log> responseToEnvelope(ReadResponse response) {
         return Flux.fromIterable(response.getEnvelopes().getBatch())
-                .map(Envelope::getLog)
-                .filter(Objects::nonNull);
+                .filter(envelope -> envelope.getLog() != null)
```
onobc, line 84: These 2 new lines do the same thing as they did before but the previous uses method refs and is cleaner IMO.

spring-cloud-deployer-cloudfoundry/src/main/java/org/springframework/cloud/deployer/spi/cloudfoundry/CloudFoundryAppDeployer.java
```diff
@@ -44,6 +44,7 @@
 import org.cloudfoundry.operations.services.BindServiceInstanceRequest;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.cloud.deployer.spi.app.AppInstanceStatus;
```
onobc, line 47: Is this in the proper order?

spring-cloud-deployer-cloudfoundry/src/main/java/org/springframework/cloud/deployer/spi/cloudfoundry/CloudFoundryAppDeployer.java
```diff
@@ -212,7 +213,14 @@ public void undeploy(String id) {
 
 	@Override
 	public String getLog(String id) {
-		return applicationLogAccessor.getLog(id, Duration.ofSeconds(this.deploymentProperties.getApiTimeout()));
+		AppStatus status = status(id);
```
onobc, line 216: Wdyt about using streams API to do this, something like:
```java
@Override
public String getLog(String id) {
  AppStatus status = status(id);
  String cfGuid = status.getInstances().values().stream()
		  .map((appInstanceStatus) -> appInstanceStatus.getAttributes().get("cf-guid"))
		  .filter(Objects::nonNull)
		  .findFirst()
		  .orElseThrow(() -> new IllegalArgumentException("Unable to find cf-guid"));
  return applicationLogAccessor.getLog(cfGuid, Duration.ofSeconds(this.deploymentProperties.getApiTimeout()));
}
```
Also, it currently does not handle the case where the guid was not found (see the task code below)

spring-cloud-deployer-cloudfoundry/src/main/java/org/springframework/cloud/deployer/spi/cloudfoundry/CloudFoundryAppDeployer.java
```diff
@@ -212,7 +213,14 @@ public void undeploy(String id) {
...
+		String cfGuid = null;
+		for(Map.Entry<String, AppInstanceStatus> appInstanceStatus : status.getInstances().entrySet()) {
+			if(appInstanceStatus.getValue().getAttributes().containsKey("cf-guid")) {
```
onobc, line 219: Might be a good idea to move these id attribute names to constant somewhere.

spring-cloud-deployer-cloudfoundry/src/main/java/org/springframework/cloud/deployer/spi/cloudfoundry/CloudFoundryTaskLauncher.java
```diff
@@ -129,10 +131,14 @@ public void destroy(String appName) {
 
 	@Override
 	public String getLog(String id) {
-		return this.applicationLogAccessor.getLog(id, Duration.ofSeconds(this.deploymentProperties.getApiTimeout()));
+		TaskStatus taskStatus = this.status(id);
+		String appCfGuid = taskStatus.getAttributes().get("app-cf-guid");
+		logger.info("Retrieving log for app-guid-id {} for task-guid-id {}", appCfGuid, id);
```
onobc, line 136: Move to debug level?

spring-cloud-deployer-cloudfoundry/src/main/java/org/springframework/cloud/deployer/spi/cloudfoundry/AbstractCloudFoundryTaskLauncher.java
```diff
@@ -145,17 +147,20 @@ private TaskStatus createErrorTaskStatus(String id) {
 	}
 
 	protected TaskStatus toTaskStatus(GetTaskResponse response) {
+		Map<String, String> attributes = new HashMap<>();
+		//retrieve CF-GUID for retrieving logs
+		attributes.put("app-cf-guid", response.getTaskRelationships().getApp().getData().getId());
```
onobc, line 152: Will this call chain ever be null (NPE?)

spring-cloud-deployer-cloudfoundry/src/main/java/org/springframework/cloud/deployer/spi/cloudfoundry/ApplicationLogAccessor.java
onobc: Also, the format of `<method-name>:<arg0>:<argN>` is not too helpful and requires you to match the code to the log to understand what they are. I am in favor of using a sentence like "Retreiving log for deploymentId:{} w/ apiTimeout:{}"

spring-cloud-deployer-cloudfoundry/src/test/java/org/springframework/cloud/deployer/spi/cloudfoundry/CloudFoundryTaskLauncherTests.java
```diff
@@ -517,6 +537,11 @@ public void testCommand() {
 		assertThat(command).isEqualTo("command-val test-command-arg-1 a=b run.id=1 run.id\\\\\\(long\\\\\\)=1 run.id\\\\\\(long=1 run.id\\\\\\)=1");
 	}
 
+	@Test
+	void getLog() {
```
onobc, line 541: We are missing test coverage for these cases:
* null entries in the log
* ensuring we are getting the task guid from the proper place (correct guid)
* negative cases where the task guid was not found/retrieved

onobc: Also, no coverage added for CF app deployer changes

onobc (changes requested):
@cppwfs looks good - I have made a few minor suggestions to take a peek at. Thanks

spring-cloud-deployer-cloudfoundry/src/test/java/org/springframework/cloud/deployer/spi/cloudfoundry/CloudFoundryAppDeployerTests.java
```diff
@@ -134,8 +136,31 @@ public void deploy() throws IOException {
...
+		assertThat(deployer.getLog("test-application-id")).isEqualTo(LOG_RESULTS);
+	}
+	@Test
+	void getLogWithNoResult() {
+		when(this.operations.applications().get(any())).thenThrow(new NoSuchElementException());
+		ApplicationLogAccessor applicationLogAccessor = mock(ApplicationLogAccessor.class);
+		CloudFoundryAppDeployer deployer = new CloudFoundryAppDeployer(this.applicationNameGenerator,
+				this.deploymentProperties,
+				this.operations, this.runtimeEnvironmentInfo, applicationLogAccessor);
+		assertThatThrownBy( () -> deployer.getLog("test-application-id"))
```
onobc, line 157: Ahhh... check this more convenient form for IAEX... 
```java
assertThatIllegalArgumentException().isThrownBy(() -> ...)
```

onobc (approved):
Thanks for the updates @cppwfs - great work!

PR #416, opened 2023-12-13
Increase log envelope limit to the max allowable.

cppwfs:
This will give users more visibility into their logs


onobc (approved):
LGMT - thanks @cppwfs

PR #422, opened 2024-01-03
K8s Task Launcher needs to be thread safe

cppwfs:
Launching task asynchronously can cause task method values to be overwritten when other threads launch at the same time.

In this solution we synchronize the launch method so that launches are thread safe

### DIFF
--- a/spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java
+++ b/spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java
@@ -235,7 +235,7 @@ public class KubernetesTaskLauncher extends AbstractKubernetesDeployer implement
 	}
 
 
-	private void launch(String appId, AppDeploymentRequest request) {
+	private synchronized void launch(String appId, AppDeploymentRequest request) {
 		Map<String, String> idMap = createIdMap(appId, request);
 		Map<String, String> podLabelMap = new HashMap<>();
 		podLabelMap.put("task-name", request.getDefinition().getName());


spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java
```diff
@@ -235,7 +235,7 @@ protected String createDeploymentId(AppDeploymentRequest request) {
 	}
 
 
-	private void launch(String appId, AppDeploymentRequest request) {
+	private synchronized void launch(String appId, AppDeploymentRequest request) {
```
onobc, line 238: I see that you sync here rather than the `public void launch(AppDeploymentRequest)` method. 

Do you mind explaining how the race condition was occurring?

cppwfs: Great question. When remote partitioning a batch job you can specify a Executor like ThreadPoolTaskExecutor and thus the partitions are launched asynchronously and each partition launches a pod using the same instance of KubernetesTaskLauncher. What I noticed is that the same pod name was being re-used on some of the launch requests and thus an exception was thrown from Fabrik8 saying pod with x name already exists.

onobc: >  What I noticed is that the same pod name was being re-used on some of the launch requests 

Got it. I am trying to identify where in the code this is happening.

corneil (approved):

cppwfs:
Refer to issue https://github.com/spring-cloud/spring-cloud-deployer/issues/423

cppwfs:
rebased Merged

PR #426, opened 2024-01-19
Allow MaximumConcurrentTasks to be unlimited (lower than zero)

surmabck:
This is a possible implementation of: https://github.com/spring-cloud/spring-cloud-dataflow/issues/5646

### DIFF
--- a/spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java
+++ b/spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java
@@ -221,7 +221,14 @@ public class KubernetesTaskLauncher extends AbstractKubernetesDeployer implement
 	}
 
 	private boolean maxConcurrentExecutionsReached() {
+		if (isMaximumConcurrentTasksUnlimited()) {
+			return false;
+		}
 		return this.getRunningTaskExecutionCount() >= this.getMaximumConcurrentTasks();
+	}
+
+	private boolean isMaximumConcurrentTasksUnlimited() {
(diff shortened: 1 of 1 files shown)


spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java
```diff
@@ -221,9 +221,16 @@ public String getLog(String id) {
 	}
 
 	private boolean maxConcurrentExecutionsReached() {
+		if (isMaximumConcurrentTasksUnlimited()) {
+			return false;
+		}
 		return this.getRunningTaskExecutionCount() >= this.getMaximumConcurrentTasks();
```
corneil, line 227: How about `return this.getMaximumConcurrentTasks() > 0 && this.getCurrentTaskExecutionCount() >= this.getMaximumConcurrentTasks();`
This would allow a value of 0 to indicate no limit since 0 would normally mean no tasks can be executed

spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java
```diff
@@ -221,9 +221,16 @@ public String getLog(String id) {
 	}
 
 	private boolean maxConcurrentExecutionsReached() {
+		if (isMaximumConcurrentTasksUnlimited()) {
```
cppwfs, line 224: Also we need a unit test.

cppwfs:
Closing due to inactivity.  If you feel that this was closed in error please comment.   Thank you!

PR #432, opened 2024-01-30
Exclude java ee imports

corneil:


spring-cloud-deployer-cloudfoundry/pom.xml
```diff
@@ -43,6 +43,12 @@
 			<groupId>org.cloudfoundry</groupId>
 			<artifactId>cloudfoundry-client-reactor</artifactId>
 			<version>${cloudfoundry-java-lib.version}</version>
+			<exclusions>
```
onobc, line 46: @corneil do you mind adding a well-known comment in each of these exclusions so we can easily find it later. 

How about `// @TODO boot3 jakarta` ?

onobc (changes requested):
@corneil LGTM. One minor ask on adding comments.

spring-cloud-deployer-cloudfoundry/pom.xml
corneil: Updated

onobc (approved):
LGTM - thanks @corneil

PR #434, opened 2024-02-05
Extend timeouts in tests for deployments.

corneil:
During local executions of tests that resolve remote artefacts the throttled download of repo.spring.io causes timeouts. 
Extending the timeouts saves people from going down rabbit holes.


onobc (approved):
LGTM - thanks @corneil

PR #437, opened 2024-02-20
[Local Deployer] [Docker] Added ability to add additional docker hosts

kuba3351:
Hello

I need to use an `host.docker.internal` alias inside docker containers deployed using the Local Deployer to reach services deployed on the host machine for development purposes. Unfortunately this alias does not working out of the box even if a container is attached to a docker network. It requires to pass an `--add-host` to the `docker run` command but Local deployer does not have ability to do it.

So, I created this PR to add ability to define additional host aliases that extends the `DockerCommandBuilder` for proper passing it to the `docker run` command.


corneil (approved):

onobc (approved):
Nice work @kuba3351 - thanks for the contribution.

PR #452, opened 2024-05-16
Add dep. mgmt. for commons-io and commons-compress

onobc:
A previous update to commons-compress 1.26.0 introduced a requirement to use commons-io 2.15.1 but there was a version 2.7.x being pulled in by maven resource module. 
This commit adds dependency management to all deployer modules so that these libs stay in sync.


pom.xml
```diff
@@ -110,6 +111,16 @@
 				<type>pom</type>
 				<scope>import</scope>
 			</dependency>
+			<dependency>
```
cppwfs, line 114: nice!

cppwfs (approved):
Thanks for adding it to dependency management!

PR #461, opened 2024-07-02
Fix cfenv detection

jvalkeal:
Doing extra catch so that we don't fail with UnsupportedClassVersionError which happens if class file is found and loaded with older jdk than cfend is compiled. In this case we assue cfenv is present.

We're testing this with spring-cloud-task sample where cfenv is added via build.


onobc (commented):
LGTM @jvalkeal - thx for the fix!

PR #463, opened 2024-07-08
[DO NOT MERGE] Node-Selector id format needs to match the other properties

cppwfs:
The id as it is used in the deployer is not correct.  The constant needed to be updated.

Do not merge until after 2.9.4 is released


corneil (commented):
The resolver needs to check against multiple relaxed names.

cppwfs:
Closing this PR in light of the PR 464 issued by @corneil

PR #467, opened 2024-07-10
Put dependency versions in properties

onobc:
This PR updates Jackson to 2.17.2 as well as moves a few version numbers into Maven properties rather than being specified in the middle of the pom.xml.


corneil (approved):
LGTM

cppwfs (approved):
LGTM!  Thank you!!!

PR #470, opened 2024-08-06
Check volume requirement on all containers.

### LINKED ISSUE #454: Anyone who get trouble using property additional container & volumeMounts
**Description:**
I just want sidecar container(additional container) with my task container.
I put property like below
```
deployer.{myapp}.kubernetes.volumes=[{name: 'promtail-config', configMap: {name: promtail-config, items: [{key: promtail.yaml, path: promtail.yaml}]}}]
deployer.{myapp}.kubernetes.additional-containers=[{name: 'promtail',image: image-path-of-promtail, ports:[{protocol: TCP,containerPort: 8080}],args: ["-config.file=/home/conf/promtail.yaml"],volumeMounts: [{name: 'promtail-config', mountPath: '/home/conf'}]}]
```
of course, in my k8s cluster there is definitely `promtail-config` yaml object.

first, I put property for `volumes`
second, I put property for `additional container`. Then, I describe info about `volumeMounts`

but I cannot launch task with this error log.
```
Message: Pod "test-task-9rq9ojo3k0" is invalid: spec.containers[1].volumeMounts[0].name: Not found: "promtail-config".
```

Did I understand k8s system badly?....

I searched about this property or example for it, Unfortunately, there is no information about `additional container's volume in SCDF`

**Release versions:**
SCDF 2.10.2

**Custom apps:**
Task

Is there anyone who had problem like my situation..?

corneil:
Volume was only checked on volume-mounts of main container. 
In the case where volumes were defined for use on init-containers or additional-containers it was not added the list of volumes.
Now the volume is checked and added after all containers are known.

Fixes #454

### DIFF
--- a/spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
+++ b/spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
@@ -250,8 +250,6 @@ public class AbstractKubernetesDeployer {
 
 		podSpec.withTolerations(this.deploymentPropertiesResolver.getTolerations(deploymentProperties));
 
-		// only add volumes with corresponding volume mounts
-
 
 		if (hostNetwork) {
 			podSpec.withHostNetwork(true);
@@ -316,6 +314,7 @@ public class AbstractKubernetesDeployer {
 			allContainers.add(initContainer);
 		}
 		allContainers.addAll(additionalContainers);
+		// only add volumes with corresponding volume mounts in any container.
 		podSpec.withVolumes(this.deploymentPropertiesResolver.getVolumes(deploymentProperties).stream()
 				.filter(volume -> allContainers.stream()
 						.anyMatch(c -> c.getVolumeMounts()
--- a/spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
+++ b/spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.deployer.spi.kubernetes;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/AbstractKubernetesDeployer.java
```diff
@@ -250,10 +251,7 @@ PodSpec createPodSpec(AppDeploymentRequest appDeploymentRequest) {
 		podSpec.withTolerations(this.deploymentPropertiesResolver.getTolerations(deploymentProperties));
 
 		// only add volumes with corresponding volume mounts
```
cppwfs, line 253: This comment needs to be moved to the new location of where we add volumes to volume mounts.

spring-cloud-deployer-kubernetes/src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployerTests.java
```diff
@@ -139,6 +143,24 @@ public void deployWithVolumesAndVolumeMounts() throws Exception {
                 new VolumeBuilder().withName("testpvc").withNewPersistentVolumeClaim("testClaim", true).build(),
                 new VolumeBuilder().withName("testnfs").withNewNfs("/test/override/nfs", null, "192.168.1.1:111").build());
     }
+    @Test
+    public void deployWithVolumesAndVolumeMountsOnAdditionalContainer() throws Exception {
+        AppDefinition definition = new AppDefinition("app-test", null);
+        Map<String, String> props = new HashMap<>();
+        props.put("spring.cloud.deployer.kubernetes.volumes", "[{name: 'config', configMap: {name: promtail-config, items: [{key: promtail.yaml, path: promtail.yaml}]}}]");
```
cppwfs, line 150: Nitpick: Could we add more volumes so that we can make sure the correct one is returned?

cppwfs (changes requested):

cppwfs:
One more request can we have a negative test?   🙏

spring-cloud-deployer-kubernetes/src/test/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesAppDeployerTests.java
corneil: done

corneil:
> One more request can we have a negative test? 🙏

Added a negative test

cppwfs (approved):
LGTM

cppwfs (approved):
LGTM Thanks for looking into this!!!!!

PR #474, opened 2024-09-02
Ensure initContainer(s) will configure args provided.

### LINKED ISSUE #473: Add support to configure 'args' fields for k8s initContainer
### Description

We need to set `args` field of initContainer. 
As far as I know, I can configure initContainer only with these fields. 
`containerName`, `imageName`, `commands`, `environmentVariables`.

Do you have any solution or do you have any plan for this future?
Thank you.

corneil:
Fixes #473

### DIFF
--- a/spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
+++ b/spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
@@ -893,7 +893,7 @@ public class KubernetesDeployerProperties {
         private String containerName;
 
         private List<String> commands;
-
+        private List<String> args;
         private List<VolumeMount> volumeMounts;
 
         /**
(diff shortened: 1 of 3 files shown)


spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java
```diff
@@ -925,6 +925,14 @@ public void setCommands(List<String> commands) {
             this.commands = commands;
         }
 
+        public List<String> getArgs() {
```
onobc, line 928: It seems like before this change we were using commands to contain both command + args. If users were previously doing `commands = "sh,-c,script2.sh"` will they now have to use `commands = "sh" + args = "-c,script2.h"` or will both variants still be supported? IOW - will this break existing configurations?

corneil: This test was just to validate the provision of args are applied as expected.

`commands` overrides the entrypoint.
`args` is added to the entrypoint.

PR #490, opened 2024-11-07
Update toLowerCase and toUpperCase to use Locale.ROOT

corneil:
Update `toLowerCase` and `toUpperCase` to use `Locale.ROOT`


onobc (approved):
Thanks @corneil - LGTM

PR #493, opened 2024-11-12
Fix issue with missing dot in property name

### LINKED ISSUE #492: MIssing dot when get getDeploymentPropertyValue
We tried to use init container with environment variable 
` deployer.<application>.kubernetes.init-container.environment-variables=env1=value1,env2=value2 `
But it does not applied.

Finally it works with
` deployer.<application>.kubernetes.init-containerenvironment-variables=env1=value1,env2=value2 `

I think there is a typo
https://github.com/spring-cloud/spring-cloud-deployer/blob/552f57fbf83f9875652c4d30b11c0fd7ee840b20/spring-cloud-deployer-kubernetes/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DeploymentPropertiesResolver.java#L656

corneil:
Updated environmentVariables usage to include a dot.

Fixes #492


cppwfs (approved):
LGTM

Issue #500, opened 2024-12-02
Ephemeral Storage Property Missing from 2.9.4 Onwards


HyoSang:
Hello,

It seems that the changes from [this commit](https://github.com/spring-cloud/spring-cloud-deployer/pull/443) are no longer present starting from the 2.9.4 release.

Currently, this feature is only available in version 2.9.3. As a result, from version 2.9.4 onwards, the ephemeral-storage property cannot be applied to pods.

Could you please investigate and restore this feature?

Thank you!

PR #501, opened 2024-12-02
Fix missing ephemeral storeage and hugepages properties.

corneil:
Adds missing ephemeral storeage and hugepages properties.

See #500


cppwfs:
Resolved by #501 
Polished Squashed Merged.

